PR #19, opened 2015-03-10
Windows

wjwwood:
This pull request provides support for building `ament_cmake` and `ament_python` packages on Windows.
It provides `.bat` versions of `.sh` helper scripts like `<package_name>__build.sh` It also provides a new workflow for CMake which uses some custom options and calls out to `msbuild` rather than `make` for building and installing.

Connects to ros2/ros2#9

### DIFF
--- a/ament_tools/build_types/ament_cmake.py
+++ b/ament_tools/build_types/ament_cmake.py
@@ -113,7 +113,7 @@ class AmentCmakeBuildType(BuildType):
             if context.symlink_install:
                 cmake_args += ['-DAMENT_CMAKE_SYMLINK_INSTALL=1']
             if IS_WINDOWS:
-                cmake_args += ['-G', 'Visual Studio 12 2013 Win64']
+                cmake_args += ['-G', 'Visual Studio 14 2015 Win64']
             if CMAKE_EXECUTABLE is None:
                 raise VerbExecutionError("Could not find 'cmake' executable")
             yield BuildAction(prefix + [CMAKE_EXECUTABLE] + cmake_args)
--- a/ament_tools/build_types/ament_cmake.py
+++ b/ament_tools/build_types/ament_cmake.py
@@ -174,7 +174,6 @@ class AmentCmakeBuildType(BuildType):
             lines.append('if exist "{0}" call "{0}"\n'.format(local_setup))
         lines.append(
             'set "CMAKE_PREFIX_PATH=%CMAKE_PREFIX_PATH%;%AMENT_PREFIX_PATH%"')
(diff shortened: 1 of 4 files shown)


dirk-thomas:
lgtm

ament_tools/build_types/ament_cmake.py
```diff
@@ -107,14 +113,38 @@ def on_build(self, context):
...
+                elif vsv == '14.0':
+                    generator = 'Visual Studio 14 2015 Win64'
+                else:
+                    raise VerbExecutionError("Unknown VS version: " + vsv)
+                cmake_args += ['-G', generator]
             if CMAKE_EXECUTABLE is None:
                 raise VerbExecutionError("Could not find 'cmake' executable")
             yield BuildAction(prefix + [CMAKE_EXECUTABLE] + cmake_args)
-        else:
+        elif not IS_WINDOWS:  # Nothing to do if False on Windows
```
dirk-thomas: I find the comment a bit confusing. After reading the code its clear what it means. What about:

    check for reconfigure where available

ament_tools/build_types/ament_cmake.py
```diff
@@ -130,10 +160,48 @@ def on_install(self, context):
...
+        lines.append('@echo off\n')
+        lines.append('if defined AMENT_TRACE_SETUP_FILES echo Inside %~0')
+        for path in context.build_dependencies:
+            local_setup = os.path.join(path, 'local_setup.bat')
+            lines.append('if exist "{0}" call "{0}"\n'.format(local_setup))
+        lines.append(
+            'set "CMAKE_PREFIX_PATH=%CMAKE_PREFIX_PATH%;%AMENT_PREFIX_PATH%"')
+        lines += ['%*']
+        lines += ['if %ERRORLEVEL% NEQ 0 exit /b %ERRORLEVEL%']
+        lines += ['if defined AMENT_TRACE_SETUP_FILES echo Leaving %~0']
```
dirk-thomas: Why use `+= [...]` half of the times and `.append(...)` the other half?

dirk-thomas:
In general when adding handling for special case (in this case Windows) I usually add the "special case" behind the code for the "common" case. Imo it makes the general code more easy to read.

In this case I would prefer:

    if NOT IS_WINDOWS:
        # do common stuff
    else:
        # do Windows specific stuff

The same for single line conditions:

    ext = '.sh' if not IS_WINDOWS else '.bat'

wjwwood:
I believe the comments have been addressed.

dirk-thomas:
lgtm

esteve:
+1

PR #21, opened 2015-03-18
Fix nosetest in virtualenv

wjwwood:
The batch CI job for OS X was not finding `nosetest3` because it was using Python3, but in a virtualenv.

This pull request changes the behavior to check for, and execute, nosetest with the current python executable, i.e. `sys.executable`.

This fixes the CI job: http://54.183.26.131:8080/job/ros2_batch_ci_osx/

### DIFF
--- a/ament_tools/build_types/ament_python.py
+++ b/ament_tools/build_types/ament_python.py
@@ -23,8 +23,6 @@ import shutil
 import stat
 import sys
 
-from osrf_pycommon.process_utils import which
-
 from ament_package.templates import configure_file
 from ament_package.templates import get_environment_hook_template_path
 from ament_package.templates import get_package_level_template_names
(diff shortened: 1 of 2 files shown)


wjwwood:
@esteve @tfoote @dirk-thomas for review.

dirk-thomas:
lgtm

wjwwood:
@dirk-thomas what do we need to do to get this to use the `ament_pep8` and `ament_pyflake`? The current test failures are line length errors and rather than updating the one-off `flake8` test in this package I think we should update it to use the other tools.

tfoote:
+1

dirk-thomas:
@wjwwood Travis needs to invoke "our" linters `ament_pep8` / `ament_pyflakes`. Since we currently have no binary packages and those packages have a lot of dependencies it is not really feasible to do use them in Travis.

If these packages would be split into a pure Python and a CMake package (like `ament_copyright` and `ament_cmake_copyright`) it would be trivial to install / use them within Travis.

wjwwood:
Ok, for now I've updated the check in this package to allow for 100 character lines.

I've created issues for fixing this properly later:

ament/ament_lint#11
ament/ament_tools#22

esteve:
+1

PR #30, opened 2015-04-08
adding the --end-with and --only-package options

wjwwood:
This pr adds two new options to the `ament build` verb, `--end-with` and `--only-package`.

Examples:

```
% ament build --start-with foo --end-with bar
% ament build --end-with bar
% ament build --only-package foo
# Which is equivalent to:
% ament build --start-with foo --end-with foo
```

### DIFF
--- a/ament_tools/verbs/build/cli.py
+++ b/ament_tools/verbs/build/cli.py
@@ -100,6 +100,15 @@ def prepare_arguments(parser, args):
         '--start-with',
         help='Start with a particular package',
     )
+    parser.add_argument(
+        '--end-with',
+        help='End with a particular package',
+    )
+    parser.add_argument(
+        '--only-package',
+        '--only',
+        help='Only process a particular package, implies --start-with <pkg> and --end-with <pkg>'
+    )
 
     # Allow all available build_type's to provide additional arguments
     for build_type in yield_supported_build_types():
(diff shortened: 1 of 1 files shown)


wjwwood:
@dirk-thomas @esteve @tfoote for review.

ament_tools/verbs/build/cli.py
```diff
@@ -100,6 +100,15 @@ def prepare_arguments(parser, args):
         '--start-with',
         help='Start with a particular package',
     )
+    parser.add_argument(
+        '--end-with',
+        help='Start with a particular package',
```
esteve: s/Start/End/

esteve:
+1

ament_tools/verbs/build/cli.py
wjwwood: Oops, good catch.

ament_tools/verbs/build/cli.py
```diff
@@ -100,6 +100,15 @@ def prepare_arguments(parser, args):
         '--start-with',
         help='Start with a particular package',
     )
+    parser.add_argument(
+        '--end-with',
+        help='End with a particular package',
+    )
+    parser.add_argument(
+        '--only-package',
+        '--only',
```
dirk-thomas: NItpick: Flip the order of the two options. Similar to `-C` / `--directory`.

ament_tools/verbs/build/cli.py
```diff
@@ -141,19 +150,53 @@ def main(opts, per_package_main=build_pkg_main):
 def print_topological_order(opts, packages):
     package_names = [p.name for _, p, _ in packages]
 
+    if opts.only_package and (opts.start_with or opts.end_with):
+        # The argprase mutually exclusive mechanism is broken for subparsers
+        # See: http://bugs.python.org/issue10680
+        # So we'll check it manually here
+        sys.exit("The --start-with and --end-with options cannot be used with "
+                 "the --only-package option.")
+
```
dirk-thomas: I would move this check inside the condition from line 168.

ament_tools/verbs/build/cli.py
```diff
@@ -190,6 +233,9 @@ def iterate_packages(opts, packages, per_package_callback):
         rc = per_package_callback(opts)
         if rc:
             return rc
+        if package.name == opts.end_with:
+            print("Stopping with package '{0}'".format(package.name))
```
dirk-thomas: I would phrase this:

    Stopped after package ...

ament_tools/verbs/build/cli.py
wjwwood: It only works with single `-` options, switching the order would change the destination name. I'll leave as is (we discussed it off-line).

wjwwood:
Addressed.

dirk-thomas:
+1

wjwwood:
I'm going to merge, @tfoote let me know if you have retroactive comments.

tfoote:
lgtm

Issue #22, opened 2015-03-18
Replace custom flake8 style check with ament_pep8 and ament_pyflakes


wjwwood:
Once ament/ament_lint#11 is resolved update this package to use the newly split up tools.

AC:

- [x] This package tests it styles using `ament_pep8` and `ament_pyflakes`.

dirk-thomas:
@esteve @tfoote @wjwwood Please review the referenced PR.

wjwwood:
+1

PR #31, opened 2015-04-10
replace custom flake8 test with ament_pep8 / ament_pyflakes invocation

dirk-thomas:
Connects to ament/ament_tools#22.


esteve:
+1

dirk-thomas:
(Before merging I have to replace the temporary branch `split_linter_packages_in_python_and_cmake` in the travis file.)

.travis.yml
```diff
@@ -3,19 +3,33 @@ python:
...
+  - python setup.py install
+  - cd ../..
+  - cd ament_lint/ament_pyflakes
+  - python setup.py install
+  - cd ../..
 script:
   - python setup.py nosetests -s
+  - ament_copyright ament_tools scripts test
+  - ament_pep8 ament_tools scripts test
+  - ament_pyflakes ament_tools scripts test
```
wjwwood: I would prefer these to be part of the nose tests, as the flake8 test was before. Why did you do it this way instead?

wjwwood:
+1

PR #32, opened 2015-04-15
add --no-build option to test and test_pkg

wjwwood:
This option will skip trying to build the package before testing. This should really speed up our build farm jobs which are basically building twice for no reason. Most packages just do a quick build check and install check, but it still takes time. Others actually rebuild, for why I don't know, but it may be because the files get installed again and therefore the are touched.

@dirk-thomas @esteve @tfoote for review.

### DIFF
--- a/ament_tools/verbs/build_pkg/cli.py
+++ b/ament_tools/verbs/build_pkg/cli.py
@@ -160,11 +160,19 @@ def add_arguments(parser):
         help='Flags to be passed to make by build types which invoke make'
     )
     parser.add_argument(
+        '--skip-build',
+        action='store_true',
+        default=False,
+        help='Skip the build step (this can be used when installing or '
+             'testing and you know that the build has successfully run)',
+    )
+    parser.add_argument(
         '--skip-install',
         action='store_true',
         default=False,
         help='Skip the install step (only makes sense when install has been '
-             'done before using symlinks and no new files have been added)',
+             'done before using symlinks and no new files have been added or '
+             'when testing after a successful install)',
     )
     parser.add_argument(
         '--symlink-install',
@@ -279,10 +287,11 @@ def run(opts, context):
 
     pkg_name = context.package_manifest.name
 
-    # Run the build command
-    print("+++ Building '{0}'".format(pkg_name))
-    on_build_ret = build_type_impl.on_build(context)
-    handle_build_action(on_build_ret, context)
+    if not opts.skip_build:
+        # Run the build command
+        print("+++ Building '{0}'".format(pkg_name))
+        on_build_ret = build_type_impl.on_build(context)
+        handle_build_action(on_build_ret, context)
 
     if not opts.skip_install:
         # Run the install command


ament_tools/verbs/test/cli.py
```diff
@@ -46,6 +46,10 @@ def prepare_arguments(parser, args):
         default=False,
         help='Stop after tests with errors or failures',
     )
+    parser.add_argument(
+        '-n', '--no-build', dest='build_before_test', default=True, action='store_false',
+        help="assumes package is built, only running the tests"
```
dirk-thomas: To match other arguments I would suggest:

    help='Skip building package, assuming it has been built before'

ament_tools/verbs/test/cli.py
```diff
@@ -46,6 +46,10 @@ def prepare_arguments(parser, args):
...
+    )
```
dirk-thomas: The argument should only be specified once in the code and reused the second time to avoid deviation over time.

ament_tools/verbs/test/cli.py
```diff
@@ -46,6 +46,10 @@ def prepare_arguments(parser, args):
         default=False,
         help='Stop after tests with errors or failures',
     )
+    parser.add_argument(
+        '-n', '--no-build', dest='build_before_test', default=True, action='store_false',
```
dirk-thomas: I would suggest `--skip-build` since there is also an argument `--skip-install`.

We could also have this argument on the `build` command. When specified it would only perform the install step then.

ament_tools/verbs/test_pkg/cli.py
```diff
@@ -21,15 +21,27 @@
...
+
+
 def main(opts):
     opts.build_tests = True
     context = build_pkg_get_context(opts)
-    rc = build_pkg_run(opts, context)
-    if rc:
-        return rc
+    if opts.build_before_test:
+        rc = build_pkg_run(opts, context)
```
dirk-thomas: This function does not only perform the build but also the install step. Since there is also an argument `--skip-install` I would expect `--no-build` to only skip the build.

ament_tools/verbs/test_pkg/cli.py
```diff
@@ -21,15 +21,27 @@
...
+        if rc:
+            return rc
+    else:
+        print("Skipping build before test, because --no-build was passed.")
```
dirk-thomas: Please use single quotes where possible - or wrap `--no-build` in single quotes for better separation.

tfoote:
+1

ament_tools/verbs/test/cli.py
esteve: +1 to `--skip-build`

esteve:
+1

wjwwood:
Ok, I've pushed an entirely different commit which just adds a `--skip-build` option to `build_pkg`. So now our build farm can just do:

```
$ ./src/ament/ament_tools/scripts/ament.py test src --skip-build --skip-install
```

After building of course.

wjwwood:
@dirk-thomas @esteve @tfoote can I get a new review?

tfoote:
+1

esteve:
+1

dirk-thomas:
+1

Do you have as specific build log where it rebuild stuff on the second run? That would indicate that some part in our build process is buggy and we might want to investigate that further...

wjwwood:
All of the current build farm builds do some rebuilding:

http://54.183.26.131:8080/job/ros2_batch_ci_linux/13/consoleFull

For example, if you search for `TwoInts_ServiceTypeSupport_Introspection.cpp.o`, it occurs twice.

wjwwood:
I opened a ticket to track this: https://github.com/ros2/ros2/issues/40

I think it will take some time to figure out what is being done redundantly build to build, and to then dig into the potential causes of those items and find solutions. I don't think I have time to do it right now, but it might be something someone can pick up in their spare time.

Even if there was no rebuilding build to build, these options are probably a good idea on the farm since it will skip the make check and the redundant install/install check.

PR #33, opened 2015-06-02
add cmake build_type support

wjwwood:
This way we can use the cmake build type with the `uncrustify` package. This allows us to avoid passing ament specific CMake variables to packages which doesn't use them, like `uncrustify`.

Connects to ros2/ros2#53

### DIFF
--- a/ament_tools/build_types/ament_cmake.py
+++ b/ament_tools/build_types/ament_cmake.py
@@ -16,32 +16,22 @@
 
 import os
 
-from ament_tools.build_type import BuildAction
-from ament_tools.build_type import BuildType
-
 from ament_tools.context import ContextExtender
 
 from ament_tools.helper import extract_argument_group
 
-from ament_tools.build_types.cmake_common import CMAKE_EXECUTABLE
 from ament_tools.build_types.cmake_common import cmakecache_exists_at
-from ament_tools.build_types.cmake_common import get_visual_studio_version
-from ament_tools.build_types.cmake_common import has_make_target
-from ament_tools.build_types.cmake_common import MAKE_EXECUTABLE
 from ament_tools.build_types.cmake_common import makefile_exists_at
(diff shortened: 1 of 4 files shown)


tfoote:
+1

dirk-thomas:
The diff is almost impossible to read. I would suggest to split the change in two separate commits. The first one only copying the `ament_cmake.py` file to `cmake.py` (without any modifications) and the second one then doing the actual modification in both files (and others).

dirk-thomas:
If a CMake build type is extracted from the ament_cmake then also the command line arguments (`--force-ament-cmake-configure`, `--ament-cmake-args`) should be changed.

wjwwood:
> If a CMake build type is extracted from the ament_cmake then also the command line arguments (--force-ament-cmake-configure, --ament-cmake-args) should be changed.

I'm not sure I understand this comment. There are now four options: `--force-ament-cmake-configure`, `--force-cmake-configure`, `--ament-cmake-args`, and `--cmake-args`.

I thought of making them something like `--plain-cmake-args`, since it would reasonable for users to assume that `--cmake-args` also apply to `ament_cmake` packages, but they do not currently.

wjwwood:
Ok, I refactored the commits.

dirk-thomas:
I don't see a strong need to have every combination. I would be fine with having:
* `--force-cmake-configure` and `--cmake-args` which would affect any CMake package (also the ament ones)
* `--ament-cmake-args` to pass specific CMake flags to ament packages

wjwwood:
I agree, that's why I just left it at one for each. Your proposed way would mean less arguments being passed when you want to send an arg to all cmake instances, including ament ones. I can change it to be that.

wjwwood:
`--cmake-args` now affects both.

wjwwood:
Pushed another commit.

dirk-thomas:
LGTM

PR #35, opened 2015-07-15
implement uninstall verb

dirk-thomas:
Related to ament/ament_cmake#13

### DIFF
--- a/ament_tools/verbs/uninstall/cli.py
+++ b/ament_tools/verbs/uninstall/cli.py
@@ -1,0 +1,125 @@
+# Copyright 2015 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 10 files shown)


esteve:
+1

ament_tools/verbs/uninstall/cli.py
```diff
@@ -0,0 +1,125 @@
+# Copyright 2014 Open Source Robotics Foundation, Inc.
```
wjwwood: 2015?

wjwwood:
Probably need to mention both `on_test` and `on_uninstall` in the docs for the base class:

https://github.com/ament/ament_tools/blob/master/ament_tools/build_type.py#L95

Otherwise +1

ament_tools/verbs/uninstall/cli.py
dirk-thomas: Fixed.

dirk-thomas:
I updated `BuildType` to also implement the `on_test` and `on_uninstall` methods with `raise NotImplementedError`.

dirk-thomas:
I will go ahead and merge this. If anybody has more comment please feel free to add them afterwards.

PR #36, opened 2015-07-22
refactor generation of prefix level setup files

dirk-thomas:
Before the prefix level setup files were generated by ament_cmake packages and the tool only generated them for Python packages. With this patch they are always generated by the tool.

The second commit adds support for the symlinked install option for installed by this function.


wjwwood:
+1

PR #37, opened 2015-07-30
support gtest result files, improve information about skipped files

dirk-thomas:
The output of `ament test_results` shows no error in http://ci.ros2.org/job/ros2_batch_ci_osx/53/console since it silently ignores the gtest result file.

This patch adds support for result files with a `testsuites` root tag and also outputs the files skipped due to different root tags when being invoked with `--verbose`.


wjwwood:
+1

PR #38, opened 2015-07-30
suppress installation of Python dependencies

dirk-thomas:
Fixes ros2/examples#41.

Connects to ros2/examples#41


tfoote:
This suppresses the auto installation of dependencies, and the builds pass since it's python and just copying files. And osrf_pycommon has a fallback mode so I can't easily test the results failing.

I've confirmed that building from source now works offline. 

+1

PR #40, opened 2015-07-30
Fix make args

dirk-thomas:
Before `ament --make-args -j1` resulted in an argparse error:

    ament build_pkg: error: argument --make-flags: expected one argument


wjwwood:
+1

PR #41, opened 2015-08-13
prepend AMENT_PREFIX_PATH to CMAKE_PREFIX_PATH

dirk-thomas:
With the recent renaming of message packages to have the same name as ROS 1 packages this is the more reasonable behavior.


wjwwood:
+1

PR #48, opened 2015-09-02
invoke CTest with dashboard test summary

dirk-thomas:
http://ci.ros2.org/job/ros2_batch_ci_windows/358/

### DIFF
--- a/ament_tools/build_types/ament_python.py
+++ b/ament_tools/build_types/ament_python.py
@@ -115,7 +115,7 @@ class AmentPythonBuildType(BuildType):
             'test', context, additional_lines=additional_lines)
         xunit_file = os.path.join(
             context.build_space, 'test_results',
-            context.package_manifest.name, 'nosetests.xml')
+            context.package_manifest.name, 'nosetests.xunit.xml')
         if not os.path.exists(os.path.dirname(xunit_file)):
             os.makedirs(os.path.dirname(xunit_file))
         assert NOSETESTS_EXECUTABLE, 'Could not find nosetests'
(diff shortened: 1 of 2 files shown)


dirk-thomas:
The third commit adds the result file type as a second extension to the filename.

jacquelinekay:
I am reviewing this right now.

jacquelinekay:
Can you explain what's happening in the latest Jenkins job for the `gtest_location` branches for Windows?

jacquelinekay:
Nevermind, ros2/ros2/issues/106 explains it.

jacquelinekay:
Hmm, the double-reporting issue really bothers me but I don't know enough about these testing frameworks to suggest a fix.

Is there a way to get Jenkins to only the report failed/not run CTest results? Or is this not enough information/not a complete representation of the test suite, and why?

dirk-thomas:
Well, you kind of need to be able to see all tests. Otherwise you don't know if a certain test was actually run and what its output was.

jacquelinekay:
Ok, what I meant to ask is: why can't Jenkins just report the CTest dashboard and ignore the Gtest/Jtest results? It seems to me that the CTest results are a superset of the test results we're interested in, but I could be wrong (since again, I don't know much about the features of the different frameworks).

dirk-thomas:
We could only use the CTest result file. But that would hide all all GTest / Nosetests specific "sub-tests" from the user. It would only report a single passed / failed test for a nosetests invocation.

jacquelinekay:
Got it. I guess when it comes to tests more information is always better
than less.

+1

PR #51, opened 2015-09-09
fix uninstalling directories

dirk-thomas:
When the `install.log` contains directory names it failed before to uninstall the package (e.g. for `ament_tools` itself).


jacquelinekay:
+1, tested by installing and uninstalling my ament workspace

PR #53, opened 2015-09-15
add bash and zsh completion

dirk-thomas:


jacquelinekay:
Nice, tab completion is great. I expected to be able to tab complete after installing ament_tools with this branch and sourcing my setup.bash, since this is ROS 1 behavior, but I had to source `ament_completion/bash` separately. Are scripts installed to the `environment` folder supposed to be optional add-ons to the environment, or will they eventually get incorporated into the setup script?

dirk-thomas:
This change needs ament/ament_package#21 to automatically source the environment hook. Sorry, I forgot to mention that detail :wink:

jacquelinekay:
hah, got it. +1

PR #54, opened 2015-10-09
add --ctest-args option, use ctest to run tests on non-windows

dirk-thomas:
Test results are the same after as well as before this patch:
* http://ci.ros2.org/job/ros2_batch_ci_linux/409/testReport/
* http://ci.ros2.org/job/ros2_batch_ci_linux/408/testReport/

* http://ci.ros2.org/job/ros2_batch_ci_osx/306/testReport/
* http://ci.ros2.org/job/ros2_batch_ci_osx/305/testReport/

### DIFF
--- a/ament_tools/build_types/cmake.py
+++ b/ament_tools/build_types/cmake.py
@@ -57,14 +57,23 @@ class CmakeBuildType(BuildType):
             default=[],
             help="Arbitrary arguments which are passed to all CMake projects. "
                  "Argument collection can be terminated with '--'.")
+        parser.add_argument(
+            '--ctest-args',
+            nargs='*',
+            default=[],
+            help="Arbitrary arguments which are passed to all CTest invocations. "
+                 "The option is only used by the 'test*' verbs. "
+                 "Argument collection can be terminated with '--'.")
 
     def argument_preprocessor(self, args):
         # The CMake pass-through flag collects dashed options.
         # This requires special handling or argparse will complain about
         # unrecognized options.
         args, cmake_args = extract_argument_group(args, '--cmake-args')
+        args, ctest_args = extract_argument_group(args, '--ctest-args')
         extras = {
             'cmake_args': cmake_args,
+            'ctest_args': ctest_args,
         }
         return args, extras
 
@@ -75,6 +84,7 @@ class CmakeBuildType(BuildType):
             force_cmake_configure = True
         ce.add('force_cmake_configure', force_cmake_configure)
         ce.add('cmake_args', options.cmake_args)
+        ce.add('ctest_args', options.ctest_args)
         return ce
 
     def on_build(self, context):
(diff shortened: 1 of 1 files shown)


ament_tools/build_types/cmake.py
```diff
@@ -167,12 +177,15 @@ def _common_cmake_on_test(self, context, build_type):
...
-                yield BuildAction(cmd)
+                if CTEST_EXECUTABLE is None:
+                    raise VerbExecutionError("Could not find 'ctest' executable")
+                # invoke CTest directly in order to pass arguments
+                yield BuildAction(prefix + [
+                    CTEST_EXECUTABLE,
+                    # show all test output
+                    '-V',
+                    '--force-new-ctest-process'] +
+                    context.ctest_args)
```
wjwwood: Is the `make test` target nothing more than invoking `ctest`? Do you not need to tell `ctest` where the tests are? Why not continue to use the `ARGS` method?

dirk-thomas: Yes, but using the `test` target of `make` relies on the fact that the generated code internally uses the `ARGS` environment variable.

No, `ctest` is invoked in the build space - the same as `make test`.

On Windows we already call `ctest` directly and I think calling it directly instead of indirectly is simpler and therefore preferred.

wjwwood: > Yes, but using the test target of make relies on the fact that the generated code internally uses the ARGS environment variable.

I don't understand what you mean by that. `ctest` uses `ARGS` directly (it's a feature of `ctest`), and I don't think any of our generated code needs to do anything with it.

Either way I don't care as long as you haven't overlooked a difference between the `make test` target and invoking `ctest` directly.

wjwwood:
+1

ament_tools/build_types/cmake.py
dirk-thomas: I don't think so. `ctest` does not use the environment variable `ARGS` at all. The value of it does not have any effect in the following command:

    ARGS="--help" ctest

The `test` target in a generated `Makefile` looks like this:

    test:
    	@$(CMAKE_COMMAND) -E cmake_echo_color --switch=$(COLOR) --cyan "Running tests..."
    	/usr/bin/ctest --force-new-ctest-process $(ARGS)

dirk-thomas: With stuff like http://public.kitware.com/Bug/view.php?id=8438 potentially being added in the feature it could very well be that such things will be backed into the make target and would therefore be bypassed by invoking ctest directly. So I will remove the usage of ctest from this PR.

There is only one questions: how to handle `ARGS` then? I would argue since merging `ARGS` with custom `--ctest-args` is not feasible ignore `ARGS` all together?

wjwwood: Why can't you just do something like:

```python
os.environ['ARGS'] = os.environ.get('ARGS', '') + ' '.join(options.ctest_args)
```

dirk-thomas: Since the options can easily collide.

wjwwood: So how is `ARGS="-V" ament.py test ... --ctest-args -V` different from `ament.py test ... --ctest-args -V -V`?

dirk-thomas: That the user specifies it in one single argument rather then a command line argument and an "invisible" environment variable which might not be set right in front of this line.

dirk-thomas:
I updated the PR to only add the --ctest-args option. `ARGS` and `ctest-args` will be concatenated and if the users specifies colliding arguments its simply his fault.

PR #55, opened 2015-10-12
fix quoting of ctest args

dirk-thomas:
* http://ci.ros2.org/job/ros2_batch_ci_linux/424/
* http://ci.ros2.org/job/ros2_batch_ci_osx/334/
* http://ci.ros2.org/job/ros2_batch_ci_windows/471/


wjwwood:
+1

PR #56, opened 2015-10-23
Env hooks for plain cmake

wjwwood:
So while testing `class_loader` I found that it's dependency `console_bridge`'s shared library could not be found at runtime (when running the tests). This was because `console_bridge` has no `local_setup.*` files and no environment hooks to add it's prefix to the `PATH` and `LD_LIBRARY_PATH`, etc.

I believe this problem is usually hidden because those prefixes are added to the appropriate `*PATH*`'s by other ament packages which install to the same place.

This pull request generates env hooks and setup files for plain cmake packages picked up by `ament_tools`.

I also added a message when ament is deploying files during installation and check to make sure when installing these extra files to the install prefix that we're not overwriting something that the original cmake package installed itself. You can test this by touching a file it would generate and install, e.g. `<prefix>/share/console_bridge/local_setup.sh`, and running the install of that package.

I'm pretty sure this is the right approach, but I'd welcome feedback on the idea.

### DIFF
--- a/ament_tools/build_types/cmake.py
+++ b/ament_tools/build_types/cmake.py
@@ -241,8 +241,7 @@ class CmakeBuildType(BuildType):
         path_template_path = get_environment_hook_template_path('path' + ext)
         deploy_file(
             context, os.path.dirname(path_template_path), os.path.basename(path_template_path),
-            dst_subfolder=environment_hooks_path,
-            executable=True)
+            dst_subfolder=environment_hooks_path)
 
         environment_hooks = [path_template_path]
 
@@ -252,8 +251,7 @@ class CmakeBuildType(BuildType):
             deploy_file(
                 context,
                 os.path.dirname(library_template_path), os.path.basename(library_template_path),
-                dst_subfolder=environment_hooks_path,
-                executable=True)
+                dst_subfolder=environment_hooks_path)
             environment_hooks.append(library_template_path)
 
         # expand package-level setup files
@@ -266,7 +264,7 @@ class CmakeBuildType(BuildType):
                 context,
                 os.path.dirname(destination), os.path.basename(destination),
                 dst_subfolder=rel_destination_dir,
-                executable=True, skip_if_exists=True)
+                skip_if_exists=True)
 
     def _common_cmake_on_install(self, context):
         # Figure out if there is a setup file to source
--- a/ament_tools/build_types/cmake.py
+++ b/ament_tools/build_types/cmake.py
@@ -16,11 +16,14 @@
 
 import os
 
+from ament_package.templates import get_environment_hook_template_path
+
 from ament_tools.build_type import BuildAction
 from ament_tools.build_type import BuildType
 
 from ament_tools.context import ContextExtender
 
+from ament_tools.helper import deploy_file
 from ament_tools.helper import extract_argument_group
 
 from ament_tools.build_types.cmake_common import CMAKE_EXECUTABLE
@@ -34,6 +37,7 @@ from ament_tools.build_types.cmake_common import MSBUILD_EXECUTABLE
 from ament_tools.build_types.cmake_common import project_file_exists_at
 from ament_tools.build_types.cmake_common import solution_file_exists_at
 
+from ament_tools.build_types.common import expand_package_level_setup_files
 from ament_tools.build_types.common import get_cached_config
 from ament_tools.build_types.common import set_cached_config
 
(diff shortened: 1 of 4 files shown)


ament_tools/helper.py
```diff
@@ -222,17 +222,37 @@ def extract_argument_group(args, delimiting_option):
...
+    skip_if_exists=False
+):
+    # copy the file if not already there and identical
+    source_path = os.path.join(source_base_path, filename)
+
     # create destination folder if necessary
     destination_path = os.path.join(
         context.install_space, dst_subfolder, filename)
+    # If the file exists and we should skip if we didn't install it.
+    if os.path.exists(destination_path) and skip_if_exists:
```
dirk-thomas: `os.path.exists()` won't work if the "file" exists but is a link which points to a not existing file.

ament_tools/helper.py
```diff
@@ -222,17 +222,37 @@ def extract_argument_group(args, delimiting_option):
...
+        if (
+            not os.path.islink(destination_path) or
+            not os.path.samefile(source_path, destination_path)
+        ):
+            # Finally if the content is not the same.
+            if not filecmp.cmp(source_path, destination_path):
+                # We (probably) didn't install it and shouldn't overwrite it.
+                print("-- [ament] Skipping (would overwrite):", destination_path)
+                return
+    print("-- [ament] Deploying:", destination_path)
```
dirk-thomas: This seems to be very verbose for every file being deployed by this function.

wjwwood: CMake prints out every file it installs, why shouldn't we? It's surprising to users to find out the build tool is dropping files into the install destination without logging the act.

ament_tools/helper.py
wjwwood: Ok, I updated to check to see if it exists or if it is a link. Should that be sufficient?

dirk-thomas: Yeah, lgtm.

ament_tools/helper.py
```diff
@@ -222,17 +222,40 @@ def extract_argument_group(args, delimiting_option):
...
+    ):
+        # If the dest is not a symlink or if it is but it doesn't point to our source.
+        if (
+            not os.path.islink(destination_path) or
+            not os.path.samefile(source_path, destination_path)
+        ):
+            # Finally if the content is not the same.
+            if not filecmp.cmp(source_path, destination_path):
+                # We (probably) didn't install it and shouldn't overwrite it.
+                print("-- [ament] Skipping (would overwrite):", destination_path)
```
dirk-thomas: Please use single quotes where possible (to be consistent with the existing code).

Same below.

ament_tools/helper.py
dirk-thomas: These are the first lines outputting something verbose like `-- [ament] ...`. I understand why a message would be helpful but think it is extra long. I am not sure what a shorter message could be though (still indicating that it is ament doing something beyond the package itself).

dirk-thomas:
Beside the small comments: LGTM.

ament_tools/helper.py
wjwwood: Well, I don't know of a better one either, but if you think of something I can change it retroactively.

wjwwood:
I addressed the comments, so I'll merge so I can make progress on https://github.com/ros/class_loader/pull/28.

dirk-thomas:
The generated environment hooks point to the source space instead of the install space. See fix #59.

ament_tools/build_types/cmake.py
```diff
@@ -216,6 +220,52 @@ def on_install(self, context):
...
+        destinations = expand_package_level_setup_files(
+            context, environment_hooks, environment_hooks_path)
+        for destination in destinations:
+            rel_destination_dir = \
+                os.path.dirname(os.path.relpath(destination, context.build_space))
+            deploy_file(
+                context,
+                os.path.dirname(destination), os.path.basename(destination),
+                dst_subfolder=rel_destination_dir,
+                skip_if_exists=True)
```
dirk-thomas: The `skip_if_exists` option has a pretty bad side effect. If for any reason the template changes (for example to fix bugs of this PR) this prevents the new expanded template from the build space to be copied to the install space.

Despite the idea of this option it the files generated by this function must be redeployed the next time and replace the existing file.

wjwwood: That's true, but I also wanted to avoid overwriting any files which were installed by the plain CMake package. I think the frequency with which these template files will be updated within an already built workspace will be very uncommon. On the other hand, if someone drops a new cmake package into a workspace which also installs a `setup.sh` file and we overwrite it then it will probably cause them some grief. I think the likely hood of the later case happening is similar to the case you've described, but far more confusing to the user (who are probably not experts in the system).

It would be great to accomplish both goals with improved logic, but I didn't see a way to do it reliably. Could we remove these files just before the install step? Would that work around the issue?

In general I think we should just use it sparingly.

wjwwood: Thinking about it more, could we change it to this:

- Before installation remove any files deployed with this flag set true
- Run the `make install` target
- Error if the file we're deploying is already there

It seems like that might be a better solution, but maybe there are other drawbacks I can't foresee.

dirk-thomas: Removing the files before calling the install target sounds great. But I would propose that the installation does not error out if the package installs the file on its own. It should be possible for the package to provide its own file if it wants to.

wjwwood: Well, I think it depends on the file. For example, if it is the `local_setup.sh` file, and the cmake package replaces it with an unrelated file, then the package chaining may not work correctly. But I can make a patch that removes them first, and just says it's skipping them otherwise (current behavior).

dirk-thomas: Sounds good.

PR #58, opened 2015-10-27
pass Python executable to prefix level templates

dirk-thomas:
Connects to ament/ament_package#7


dirk-thomas:
And remove executable bit from prefix level setup files (which were not removed in 9b79ca7421672233b460ecb0cd7469729f77c17f).

wjwwood:
lgtm

PR #60, opened 2015-10-27
change how skip_if_exists is used

wjwwood:
See: https://github.com/ament/ament_tools/pull/56#discussion_r43160183

### DIFF
--- a/ament_tools/build_types/cmake.py
+++ b/ament_tools/build_types/cmake.py
@@ -247,8 +247,8 @@ class CmakeBuildType(BuildType):
                 os.path.basename(destination),
                 os.path.dirname(os.path.relpath(destination, context.build_space))
             )
-            if os.path.exists(destination_path):
-                os.unlink(destination_path)
+            if os.path.exists(destination_path) or os.path.islink(destination_path):
+                os.remove(destination_path)
 
         # Call cmake common on_install (defined in CmakeBuildType)
         for step in self._common_cmake_on_install(context):
--- a/ament_tools/build_types/cmake.py
+++ b/ament_tools/build_types/cmake.py
@@ -23,6 +23,7 @@ from ament_tools.build_type import BuildType
 
 from ament_tools.context import ContextExtender
 
+from ament_tools.helper import compute_deploy_destination
 from ament_tools.helper import deploy_file
 from ament_tools.helper import extract_argument_group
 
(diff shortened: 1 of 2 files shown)


ament_tools/build_types/cmake.py
```diff
@@ -216,6 +217,39 @@ def _common_cmake_on_test(self, context, build_type):
...
+
+        # Remove package level setup files so they can be replaced correctly either in the
+        # cmake install step or later with deploy_file(..., skip_if_exists=True)
+        for destination in destinations:
+            destination_path = compute_deploy_destination(
+                context,
+                os.path.basename(destination),
+                os.path.dirname(os.path.relpath(destination, context.build_space))
+            )
+            if os.path.exists(destination_path):
```
dirk-thomas: This is not enough if the `destination_path` is a link pointing to a non existing file.

ament_tools/helper.py
```diff
@@ -222,6 +222,10 @@ def extract_argument_group(args, delimiting_option):
     return trimmed_args, extracted_args
 
 
+def compute_deploy_destination(context, filename, dst_subfolder=''):
+    return os.path.join(context.install_space, dst_subfolder, filename)
```
dirk-thomas: This custom function only takes three arguments and passes them to `os.path.join` (except using `install_space`). Does this really have to be a separate function? I don't see a value over using `os.path.join` directly.

wjwwood: It's used from two places. I did this incase the method of building this changed, then you'd be forced to change it in both locations.

dirk-thomas: I don't see how it should change. It is constructing a path.

ament_tools/build_types/cmake.py
wjwwood: Ok, now checks for `islink` too, and I changed it to use `os.remove`.

ament_tools/helper.py
wjwwood: Well how the path is constructed could change. I saw non-trivial construction of a value, being duplicated into two places and feared that someday someone might change one but not the other, so I made it into a function.

wjwwood:
Are there any other outstanding issues?

dirk-thomas:
Beside my previous comment : lgtm

PR #61, opened 2015-10-28
add pep257 check

dirk-thomas:
Also adds missing test dependencies.

### DIFF
--- a/ament_tools/build_type.py
+++ b/ament_tools/build_type.py
@@ -18,7 +18,8 @@ from .context import ContextExtender
 
 
 class BuildAction(object):
-    """Represents an action to do at build time, either a command or a functor.
+
+    """Represent an action to do at build time, either a command or a functor.
 
     These objects are yielded from the ``on_*`` methodsin the BuildType class
     for a particular ``build_type``.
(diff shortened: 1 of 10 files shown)


dirk-thomas:
Please review.

ament_tools/build_type.py
```diff
@@ -18,7 +18,9 @@
 
 
 class BuildAction(object):
-    """Represents an action to do at build time, either a command or a functor.
+
+    """
```
tfoote: Docstrings should always start on the line with the quotes. both single and multi line https://www.python.org/dev/peps/pep-0257/#id17

wjwwood: Reverse anchor link: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

tfoote: I guess it can be on the following line too.  

> The summary line may be on the same line as the opening quotes or on the next line.

dirk-thomas: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings says: The summary line may be on the same line as the opening quotes or on the next line.

I will undo the unnecessary wraps to keep the diff small.

dirk-thomas: Fixed in 1a87a3bdf0d96d3c6c3b0d604478c3b715a363c1

tfoote:
+1

PR #62, opened 2015-10-28
update build space environment files with AMENT_TRACE_SETUP_FILES

dirk-thomas:
Enables to get trace output for the environment used for building / testing / installing packages.

### DIFF
--- a/ament_tools/build_types/cmake.py
+++ b/ament_tools/build_types/cmake.py
@@ -324,15 +324,16 @@ class CmakeBuildType(BuildType):
     def _get_command_prefix_windows(self, name, context):
         lines = []
         lines.append('@echo off\n')
-        lines.append('if defined AMENT_TRACE_SETUP_FILES echo Inside %~0')
         for path in context.build_dependencies:
             local_setup = os.path.join(path, 'local_setup.bat')
-            lines.append('if exist "{0}" call "{0}"\n'.format(local_setup))
+            lines.append(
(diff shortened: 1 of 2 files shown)


ament_tools/build_types/cmake.py
```diff
@@ -324,15 +324,16 @@ def _get_command_prefix(self, name, context):
     def _get_command_prefix_windows(self, name, context):
         lines = []
         lines.append('@echo off\n')
-        lines.append('if defined AMENT_TRACE_SETUP_FILES echo Inside %~0')
         for path in context.build_dependencies:
             local_setup = os.path.join(path, 'local_setup.bat')
-            lines.append('if exist "{0}" call "{0}"\n'.format(local_setup))
+            lines.append(
+                'if "%AMENT_TRACE_SETUP_FILES%" NEQ "" echo call "{0}"'.format(local_setup))
+            lines.append('if exist "{0}" call "{0}"'.format(local_setup))
```
wjwwood: Should the trace message for this reflect whether or not the file exists?

wjwwood:
Other than one comment, lgtm.

ament_tools/build_types/cmake.py
dirk-thomas: None of the setup files does that. So no, it always echos what is supposed to be called and only calls it when it exists.

wjwwood: Well, do what you want, but it is throwing information away that is important when debugging what is happening.

PR #65, opened 2015-11-05
only support VS2015

dirk-thomas:
While ament does not really care this will make it obvious when people try building ROS 2 with older VS versions.

http://ci.ros2.org/job/ros2_batch_ci_windows/571/


jacquelinekay:
What's going on with the failed check? The code looks fine to me.

dirk-thomas:
The patches used Python 3 specific code - I wasn't realizing that we run the tests with Python 2 too. Fixed.

esteve:
+1

PR #66, opened 2015-11-14
call flush() on pipes before invoking commands

dirk-thomas:
Otherwise depending on the context where `ament` is being invoked the command output might arrive before the message printed from within `ament_tools`.


tfoote:
lgtm

PR #67, opened 2015-11-28
use --xunit-prefix-with-testsuite-name option of upcoming nosetests version

dirk-thomas:
https://github.com/nose-devs/nose/pull/968


gerkey:
+1

PR #69, opened 2015-12-10
use tuple for __all__ to fix pep257 warning

dirk-thomas:
http://ci.ros2.org/job/ci_linux/694/consoleFull#console-section-126


wjwwood:
+1

PR #70, opened 2015-12-16
generate xml of test summary

dirk-thomas:
While this will "duplicate" the error reporting on Linux and OS X (the same way it is already on the Windows jobs) it helps to debug failing tests faster (e.g. if a test doesn't generate a result file the CTest log often contains more information).

### DIFF
--- a/ament_tools/build_types/cmake.py
+++ b/ament_tools/build_types/cmake.py
@@ -185,7 +185,10 @@ class CmakeBuildType(BuildType):
                     raise VerbExecutionError("Could not find 'make' executable")
                 cmd = prefix + [MAKE_EXECUTABLE, 'test']
                 if 'ARGS' not in os.environ:
-                    args = ['-V']
+                    args = [
+                        '-V',
+                        # verbose output and generate xml of test summary
+                        '-D', 'ExperimentalTest', '--no-compress-output']
                 elif os.environ['ARGS']:
                     args = [os.environ['ARGS']]
                 else:


ament_tools/build_types/cmake.py
```diff
@@ -185,7 +185,10 @@ def _common_cmake_on_test(self, context, build_type):
                     raise VerbExecutionError("Could not find 'make' executable")
                 cmd = prefix + [MAKE_EXECUTABLE, 'test']
                 if 'ARGS' not in os.environ:
-                    args = ['-V']
+                    args = [
+                        '-V',
+                        # verbose output and generate xml of test summary
+                        '-D', 'ExperimentalTest', '--no-compress-output']
```
gerkey: Where does `-D ExperimentalTest` get used? Is that a special ctest symbol?

gerkey:
With this change, I get much less output to console. Looks like some combination of stderr and/or stdout is hidden from the console (perhaps redirected to a file?). Is that the intent?

ament_tools/build_types/cmake.py
dirk-thomas: Yes, the new args are CTest arguments.

dirk-thomas:
The new arguments should not affect the console output. Can you please describe what you ran and what the output was.

gerkey:
Turns out that I had an older version of this branch that contained a bug that was fixed in a subsequent commit. But there was no way for me to know, even though I was watching this PR, that I needed to pull to get the newer version, because the commit with the fix was squashed.

ament_tools/build_types/cmake.py
gerkey: Following an in-person conversation with @dirk-thomas, I learned that:

* `-D ExperimentalTest` is an option to ctest to cause it to perform a kind of dashboard test that saves more information.
* `--no-compress-output` prevents that file from being archived / compressed.
* The new file can be found in `build/<pkg>/Testing/<timestamp>/Test.xml`

gerkey:
+1

PR #71, opened 2015-12-17
list_packages prints names and paths

dirk-thomas:
Before only either or was possible.


wjwwood:
lgtm. not sure why py2.7 failed.

dirk-thomas:
PyFlakes didn't like:

    [print(x) for x in list]

PR #72, opened 2016-01-15
build release on Windows

dirk-thomas:
For now hard code the `Release` configuration on Windows.

http://ci.ros2.org/job/ci_windows/918/ showing almost all Connext tests are passing.

In a follow up change:
* `ament_tools` should allow to choose the configuration
* `connext_cmake_module` needs to select the libraries matching the requested build type (ros2/rmw_connext#89)
* the tests need to use generator expressions to determine their location based on the configuration selected in VS


wjwwood:
lgtm

PR #73, opened 2016-01-25
determine VS configuration based on CMake build type

dirk-thomas:
Instead of using a fixed `configuration` on Windows it is selected by the `CMAKE_BUILD_TYPE` (either specified on the command line or retrieved from the CMake cache). The fallback is `Debug` if nothing is specified.

* http://ci.ros2.org/job/ci_windows/941/
* http://ci.ros2.org/view/packaging/job/packaging_windows/106/


wjwwood:
lgtm

dirk-thomas:
Use the Visual Studio configuration also in the `install` step: 8233c9c5ae4e93ebba5e5401dc13a597d45b0b39

PR #74, opened 2016-02-03
add option to ament build to skip packages

wjwwood:
This option is needed specifically to let us skip some leaf packages on some CI jobs on the farm.

### DIFF
--- a/ament_tools/verbs/build/cli.py
+++ b/ament_tools/verbs/build/cli.py
@@ -109,6 +109,11 @@ def prepare_arguments(parser, args):
         '--only',
         help='Only process a particular package, implies --start-with <pkg> and --end-with <pkg>'
     )
+    parser.add_argument(
+        '--skip-packages',
+        nargs='*',
+        help='List of packages to skip'
+    )
 
     # Allow all available build_type's to provide additional arguments
     for build_type in yield_supported_build_types():
@@ -158,6 +163,18 @@ def print_topological_order(opts, packages):
         sys.exit("Package '{0}' specified with --end-with was not found."
                  .format(opts.end_with))
 
+    opts.skip_packages = opts.skip_packages or []
+    nonexistent_skip_packages = []
+    for skip_package in opts.skip_packages:
+        if skip_package not in package_names:
+            nonexistent_skip_packages.append(skip_package)
+        if skip_package == opts.only_package:
+            sys.exit("Cannot --skip-packages and --only-package the same package: '{0}'."
+                     .format(skip_package))
+    if nonexistent_skip_packages:
+            sys.exit('Packages [{0}] specified with --skip-packages were not found.'
+                     .format(', '.join(nonexistent_skip_packages)))
+
     if opts.only_package:
         if opts.start_with or opts.end_with:
             # The argprase mutually exclusive mechanism is broken for subparsers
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I don't think this is needed. The much simpler approach from ros2/common_interfaces#8 is sufficient for this case.

wjwwood:
We'll merge this, but just to make the tool and CI jobs more flexible.

wjwwood:
Please review for merge.

ament_tools/verbs/build/cli.py
```diff
@@ -158,6 +163,18 @@ def print_topological_order(opts, packages):
         sys.exit("Package '{0}' specified with --end-with was not found."
                  .format(opts.end_with))
 
+    opts.skip_packages = opts.skip_packages or []
+    nonexistant_skip_packages = []
```
dirk-thomas: `nonexistent`?

ament_tools/verbs/build/cli.py
```diff
@@ -158,6 +163,18 @@ def print_topological_order(opts, packages):
...
+    for skip_package in opts.skip_packages:
+        if skip_package not in package_names:
+            nonexistant_skip_packages.append(skip_package)
+        if skip_package == opts.only_package:
+            sys.exit("Cannot --skip-packages and --only-package the same package: '{0}'."
+                     .format(skip_package))
+    if nonexistant_skip_packages:
+            sys.exit("Packages [{0}] specified with --skip-packages were not found."
```
dirk-thomas: Single quotes.

ament_tools/verbs/build/cli.py
```diff
@@ -158,6 +163,18 @@ def print_topological_order(opts, packages):
...
+                     .format(', '.join(nonexistant_skip_packages)))
```
dirk-thomas: Should this really be an error or just a warning?

ament_tools/verbs/build/cli.py
```diff
@@ -200,38 +222,39 @@ def print_topological_order(opts, packages):
...
+    opts.skip_packages = opts.skip_packages or []
     install_space_base = opts.install_space
     package_dict = dict([(path, package) for path, package, _ in packages])
     for (path, package, depends) in packages:
         if package.name == opts.start_with:
             start_with_found = True
-        if not start_with_found:
+        if not start_with_found or package.name in opts.skip_packages:
             print('# Skipping: %s' % package.name)
-            continue
```
dirk-thomas: Why remove the `continue` here and intend all of the following code in an `else` block?

wjwwood: The `else:` block is new and there is code that comes after it but before the for loop continues (the `if package.name == opts.end_with:` stuff).

ament_tools/verbs/build/cli.py
wjwwood: It is in the other options, I think it is better so it catches typos. A warning would probably be missed since it is at the beginning of the output.

ament_tools/verbs/build/cli.py
wjwwood: Fixed in 10c887c

ament_tools/verbs/build/cli.py
wjwwood: 10c887c

dirk-thomas:
LGTM

PR #75, opened 2016-02-05
improve logic to determine build type

dirk-thomas:
CMake is not consistently using the PATH type for the cache entry.

Followup of #73.

### DIFF
--- a/ament_tools/build_types/cmake.py
+++ b/ament_tools/build_types/cmake.py
@@ -231,14 +231,18 @@ class CmakeBuildType(BuildType):
                 break
         else:
             # get for CMake build type from the CMake cache
-            line_prefix = 'CMAKE_BUILD_TYPE:STRING='
+            line_prefix = 'CMAKE_BUILD_TYPE:'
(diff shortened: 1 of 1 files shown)


ament_tools/build_types/cmake.py
```diff
@@ -231,14 +231,18 @@ def _get_visual_studio_configuration(self, context):
...
                 with open(cmake_cache, 'r') as h:
                     lines = h.read().splitlines()
                 for line in lines:
                     if line.startswith(line_prefix):
-                        build_type = line[len(line_prefix):]
+                        try:
+                            index = line.index('=')
+                        except ValueError:
+                            continue
+                        build_type = line[index + 1:]
```
wjwwood: If you continue on the `ValueError`, what is the value of `index` at this line?

dirk-thomas: After `continue` this line will not be called.

wjwwood: I see.

wjwwood:
+1

PR #77, opened 2016-02-18
add support for py_modules with --symlink-install

wjwwood:
fixes ros2/examples#90

### DIFF
--- a/ament_tools/build_types/ament_python.py
+++ b/ament_tools/build_types/ament_python.py
@@ -318,6 +318,9 @@ class AmentPythonBuildType(BuildType):
         items = ['setup.py']
         # add all first level packages
         items += [p for p in context['setup.py']['packages'] if '.' not in p]
+        # relative python-ish paths are allowed as entries in py_modules, see:
+        # https://docs.python.org/3.5/distutils/setupscript.html#listing-individual-modules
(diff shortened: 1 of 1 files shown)


esteve:
+1

esteve:
Sorry, when I reviewed this I didn't erase my workspace. Fix in #78

ament_tools/build_types/ament_python.py
```diff
@@ -318,6 +318,9 @@ def _install_action_python(self, context):
         items = ['setup.py']
         # add all first level packages
         items += [p for p in context['setup.py']['packages'] if '.' not in p]
+        # relative python-ish paths are allowed as entries in py_modules, see:
+        # https://docs.python.org/3.5/distutils/setupscript.html#listing-individual-modules
+        items += [re.sub(r'\.', os.path.sep, p) for p in context['setup.py']['py_modules']]
```
dirk-thomas: Why does this substitution has to be a regex? Wouldn't a plain string replace work too and be much more efficient?

wjwwood: That's true, I was thinking it would be a more complicated substitution but ended up simplifying it. I'll open a pr to change it.

dirk-thomas: @wjwwood ping

dirk-thomas: Follow up #80

PR #78, opened 2016-02-18
Handle empty py_modules gracefully

esteve:
Parse setup correctly if it doesn't have a py_modules key.

Connects to #77


wjwwood:
Thanks, I'm not sure why I didn't catch that either.

wjwwood:
+1

PR #82, opened 2016-02-27
Fix autocomplete for build_pkg/test_pkg options

dhood:
Completion for build_pkg verb was showing the options for build.

Before:
```
$ ament build_pkg -
--ament-cmake-args             --end-with                     --isolated                     --skip-install
--build-space                  --force-ament-cmake-configure  --make-flags                   --start-with
--build-tests                  --force-cmake-configure        --only-package                 --symlink-install
--cmake-args                   --install-space                --skip-build                   -C
```

After:
```
$ ament build_pkg --
--build-space      --build-tests      --install-space    --make-flags       --skip-build       --skip-install     --symlink-install
```

### DIFF
--- a/completion/ament-completion.bash
+++ b/completion/ament-completion.bash
@@ -28,7 +28,7 @@ _ament()
     COMPREPLY=( $( compgen -P "-DCMAKE_BUILD_TYPE=" -W "Debug MinSizeRel None Release RelWithDebInfo" -- "${cur:19}" ) )
   else
     if [[ "${COMP_WORDS[@]}" == *" build_pkg"* || "${COMP_WORDS[@]}" == *" test_pkg"* ]] ; then
-      COMPREPLY=($(compgen -W "--build-space --build-tests --install-space --make-flags --skip-build --skip-install --symlink-install" -- ${cur}))
+      COMPREPLY=($(compgen -W "--ament-cmake-args --build-space --build-tests --cmake-args --ctest-args --force-ament-cmake-configure --force-cmake-configure --install-space --make-flags --skip-build --skip-install --symlink-install" -- ${cur}))
     elif [[ "${COMP_WORDS[@]}" == *" build"* || "${COMP_WORDS[@]}" == *" test"* ]] ; then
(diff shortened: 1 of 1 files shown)


wjwwood:
Thanks for the pr @dhood!

One concern I had is that this might break the completion with `ament build_pkg --ament-cmake-args <tab>` when you're in a `cmake` or `ament_cmake` package. For example you can go into `src/ros2/rcl/rcl` and execute `ament build_pkg -h` and you should get some extra options for `--force-ament-cmake-configure` and `--ament-cmake-args`. However, I think that overmatching completion arguments is worse than missing these arguments which are dynamically added based on context. So I don't think you need to change anything with this pr.

dhood:
And thanks for the review @wjwwood 

I agree that the arguments `--force-ament-cmake-configure` and `--ament-cmake-args` are added dynamically based on context. 

But I feel that the completion of those options is just caused by them being in the list at https://github.com/ament/ament_tools/blob/release-alpha4/completion/ament-completion.bash#L40

For example, you can do `ament build . --ctest-args`... as the argument is listed in `ament build -h`, but `ctest-args` does complete because it's not in the aforementioned list. As such I'm not actually sure that the completion of options is being done based on context at the moment.

If you agree, then perhaps just adding more arguments to the list at https://github.com/ament/ament_tools/pull/82/files#diff-072aa2f20d47f0ccb9c37374765c1a5aR31 would make it equivalent to `build`'s completion as it stands.

wjwwood:
Hmm, I think it's fine as is.

Long term we could consider something like https://argcomplete.readthedocs.org/en/latest/, but I've been avoiding cosmetic work like this since I want to refactor `catkin_tools` and `ament_tools` so they reuse as much as possible. That process will likely touch a lot of things, so I've been sticking to necessary changes here until that's done. On the other hand, implementing https://argcomplete.readthedocs.org/en/latest/ might tell us whether or not it's a good idea to adopt it in our other tools (a good thing to prototype).

I'm going to go ahead and merge this. Others can comment and we can fix it up retroactively if there are more comments.

dhood:
yep - definitely understand that polishing isn't really a priority at the moment!

PR #87, opened 2016-04-05
update style to follow D211

dirk-thomas:
https://hg.python.org/peps/rev/9b715d8246db

Which is actually being checked for by newer versions of `pep257` / `pydocstyle`.


wjwwood:
+1

PR #92, opened 2016-04-29
use CTest BUILD_TESTING

dirk-thomas:
Connect to ament/ament_cmake#68

### DIFF
--- a/ament_tools/build_types/ament_cmake.py
+++ b/ament_tools/build_types/ament_cmake.py
@@ -93,8 +93,7 @@ class AmentCmakeBuildType(CmakeBuildType):
         # Calculate any extra cmake args which are not common between cmake build types
         extra_cmake_args = []
         if should_run_configure:
-            if context.build_tests:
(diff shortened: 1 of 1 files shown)


ament_tools/build_types/ament_cmake.py
```diff
@@ -93,8 +93,7 @@ def on_build(self, context):
         # Calculate any extra cmake args which are not common between cmake build types
         extra_cmake_args = []
         if should_run_configure:
-            if context.build_tests:
-                extra_cmake_args += ["-DAMENT_ENABLE_TESTING=1"]
+            extra_cmake_args += ["-DBUILD_TESTING=%d" % int(context.build_tests)]
```
wjwwood: This is a bit of a behavior change (previously building with `--build-tests` and then omitting it would not "undo" the building of tests), but it looks ok to me.

dirk-thomas: The problem is that before we didn't know when it was intentional and when it was not. I think the new behavior is more consistent - you get what you specify.

PR #93, opened 2016-05-02
generate ament_prefix_path hook

dirk-thomas:
With this patch even a pure CMake package will contribute its install space to the AMENT_PREFIX_PATH. Otherwise artifacts of pure CMake packages can't be found.

Depends on ament/ament_package#39.


tfoote:
+1

PR #95, opened 2016-05-20
[cmake] follow add_subdirectory/subdirs for deps

wjwwood:
Currently we only look for dependencies in the `CMakeLists.txt` in the root of a cmake package's directory. This pr makes it so that it also follows `add_subdirectory` and `subdirs` calls in cmake to other `CMakeLists.txt`.


wjwwood:
Closing in favor of https://github.com/ament/ament_tools/pull/96

PR #96, opened 2016-05-21
[cmake] check all CMakeLists.txt for deps

wjwwood:
Currently we only look for dependencies in the `CMakeLists.txt` in the root of a cmake package's directory. This pr makes it so that it also looks at all other `CMakeLists.txt` files in the project's folder.

### DIFF
--- a/ament_tools/package_types/cmake.py
+++ b/ament_tools/package_types/cmake.py
@@ -46,8 +46,17 @@ def extract_data(cmakelists):
     if not data['name']:
         raise RuntimeError("Failed to extract project name from '%s'" % cmakelists)
 
-    build_depends = extract_build_dependencies(content)
-    data['build_depends'] = [Dependency(name) for name in build_depends]
+    # Extract build dependencies from all CMakeLists.txt in the project folder.
+    build_depends_content = content
+    for root, dirs, files in os.walk(os.path.dirname(cmakelists)):
+        for name in files:
+            if name.endswith('CMakeLists.txt'):
+                with open(os.path.join(root, name), 'r') as h:
+                    build_depends_content += h.read()
+    build_depends_content = remove_cmake_comments(build_depends_content)
+    build_depends = extract_build_dependencies(build_depends_content)
+    # Convert to Dependency objects, exclude self references.
+    data['build_depends'] = [Dependency(name) for name in build_depends if name != data['name']]
 
     return data
 


wjwwood:
We decided to go ahead with this as a workaround in https://github.com/ros2/ros2/issues/223. So it needs a review.

jacquelinekay:
I started CI for this to see if it fixes the FastRTPS build. If it does I'd like to merge it ASAP so that we have reliable CI that works by default for the days leading up to the release.

http://ci.ros2.org/job/ci_linux/1337 (!)
http://ci.ros2.org/job/ci_osx/1048/
http://ci.ros2.org/job/ci_windows/1352

jacquelinekay:
+1

wjwwood:
The Linux nightly job was already run on it too: https://github.com/ros2/ros2/issues/223#issuecomment-220746488

The Linux job finished, and the OS X/Windows jobs you started are well past failing on the FastRTPS build.

I'll go ahead and merge this.

dirk-thomas:
Wouldn't it be better to at least check if a subfolder is being added before considering the content of the CMakeLists.txt file?

wjwwood:
> Wouldn't it be better to at least check if a subfolder is being added before considering the content of the CMakeLists.txt file?

 What does "is being added" mean?

dirk-thomas:
> > Wouldn't it be better to at least check if a subfolder is being added before considering the content of the CMakeLists.txt file?

> What does "is being added" mean?

CMake does not automatically process all CMakeLists.txt files. Each subdirectory needs to be added explicitly using the [add_subdirectory](https://cmake.org/cmake/help/v3.5/command/add_subdirectory.html) command.

wjwwood:
@dirk-thomas I think you need to read https://github.com/ros2/ros2/issues/223 and https://github.com/ament/ament_tools/pull/95.

https://github.com/ament/ament_tools/pull/95 does exactly this, but that's not sufficient to close https://github.com/ros2/ros2/issues/223 by itself.

dirk-thomas:
I didn't read the closed PR. I see why this is done the way it is then. I would argue that this is a pretty bad hack to make FastRTPS build. Imo this should be reverted asap and addressed upstream. Is there a ticket which tracks this since I don't think ament should continue to do this?

wjwwood:
I think the entire process of extracting the dependencies from the `CMakeLists.txt` is a fragile heuristic. It could break if the dependency is found with `pkg-config` rather than `find_package`, or if a project uses different names for the `project_name(...)` call and the generated `...Config.cmake` file, or one package expects a script to be on the path and never even does `find_package(other_package)`.

Also, I don't think we shouldn't be dictating to FastRTPS how to organize their code. We can suggest that what they're doing is strange and ask them to change it, but it's not their job to make their package build in our tool when it does build with normal CMake. I don't think it *must* be resolved upstream.

I think the right solution is to place a `package.xml` in these packages, either upstream or during our build/checkout process, so their dependencies can be explicitly declared. Finer control will be necessary anyways when we start to install dependencies (like boost) automatically.

There is no ticket to track the removal of this change. I presented three options in https://github.com/ros2/ros2/issues/223 and we discussed them in the team meeting. We decided to take this option and to let eProsima know that the way they're building the examples is odd and it caused us some issues (I haven't done that just yet). Our priority was to get the jobs turning over again before the release. I'd recommend raising your concerns on https://github.com/ros2/ros2/issues/223 and reopening that issue.

PR #98, opened 2016-06-07
add --retest-until-fail and --retest-until-pass options

dirk-thomas:
`--retest-until-fail N` will retry running all tests up to N times if all tests pass.

`--retest-until-pass N` will retry all (or for CTest only the failed) tests up to N times if any test fails.

Connect to ros2/ros2#224

### DIFF
--- a/ament_tools/build_types/cmake.py
+++ b/ament_tools/build_types/cmake.py
@@ -201,6 +201,8 @@ class CmakeBuildType(BuildType):
                 else:
                     args = []
                 args += context.ctest_args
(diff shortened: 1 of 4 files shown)


wjwwood:
> Connect to #224

Which repository is that supposed to be linked to?

dirk-thomas:
ros2/ros2#224

wjwwood:
Other than possibly putting the iteration number out of the target number in the retry messages, +1.

dirk-thomas:
Good point - updated.

PR #99, opened 2016-06-09
Added support for Gradle

esteve:
WARNING, ACHTUNG, AVISO, BEEP BEEP MORP!!!! Not really meant for merging, I'd just like some feedback.

So I've been working on adding support for Gradle to Ament, but I don't know if this is the correct approach. I wondered if there's a way to add Gradle in a modular way without actually changing Ament itself, but I couldn't figure out how.

An example which uses this and Gradle to build an .apk for Android can be found at https://github.com/esteve/ros2_talker/ (see https://github.com/esteve/ros2_talker/blob/master/build.gradle for the Gradle file)

Thanks!

:hamburger:


dirk-thomas:
The build types are registered through entry points (https://github.com/ament/ament_tools/blob/51add3f0c449db97d63dc32a99c09fab495af728/setup.py#L51-L55). This can happen within this package as well as in a separate package. Once `ament_tools` and `ament_build_type_gradle` (or whatever name) has been built the command line tool `ament` will pick up all entry points and support all available build types.

esteve:
@dirk-thomas cool, thank you so much for the explanation! So, I just need to create an new package (that uses `ament_python` as a build type) and add the following to the `setup.py`, and ament will add it as a supported build type, that's fantastic!

```
entry_points={
        'ament.build_types': [
            'ament_gradle = ament_build_type_gradle:AmentGradleBuildType',
        ],
```

Do I then have to add `ament_build_type_gradle` as a buildtool dependency too to any package that I want to build with `ament_gradle`, is that correct?

dirk-thomas:
> Do I then have to add ament_build_type_gradle as a buildtool dependency too to any package that I want to build with ament_gradle, is that correct?

Yes, that would be the cleanest. If you have built `ament_build_type_gradle` in another workspace it will work anyway.

esteve:
It worked, thanks! I moved this to a separate package and did as you said. I have to build ament first and then use the ament tool from the install space, instead of doing it all in one pass, but it's no big deal. Thanks!

dirk-thomas:
Can you please add a link to the new package so that when people find this ticket can find it.

esteve:
It's not public yet, but I already planned to link to it when I release it.

PR #100, opened 2016-06-09
Windows multiprocessing builds

codebot:
This PR will pass the /MP switch through the environment to CL.EXE so that it will do intra-process parallel builds. This appears to make the windows builds go about twice as fast. It will only happen when the -j1 flag is passed, just like the current logic only passes the /m (inter-process parallel build) flag when -j1 is supplied.


ament_tools/build_types/cmake.py
```diff
@@ -167,8 +167,9 @@ def _common_cmake_on_build(self, should_run_configure, context, prefix, extra_cm
             solution_file = solution_file_exists_at(
                 context.build_space, context.package_manifest.name)
             cmd = prefix + [MSBUILD_EXECUTABLE]
-            if '-j1' in context.make_flags:
+            if not '-j1' in context.make_flags:
                 cmd += ['/m']
+                os.environ['CL'] = '/MP'
```
dirk-thomas: Changing `os.environ` globally might become a problem. It would be better if this would only be set for this cmd / action.

codebot: That sounds better, but I don't know how to do that in this codebase.

dirk-thomas: As always Windows is not making it easier. While on other platforms you can just call `VAR=VALUE cmd` in Windows this requires some extra hoops. This function (https://github.com/ament/ament_tools/blob/51add3f0c449db97d63dc32a99c09fab495af728/ament_tools/build_types/cmake.py#L389) generates a batch file which sets up some environment and then calls whatever was passed to it (`%*`).

So either this file gets extended to contain `CL` or another file setting only `CL` is generated which is then nested.

codebot: I'm not sure how to pass the resulting batch file out of this function into whoever called us. While I agree that setting the `CL` environment variable globally on Windows could theoretically cause problems if we start using this for other purposes, I don't see any conflicts in the foreseeable future, and would propose that we merge as-is and deal with any future conflicts with `CL` if/when they arise.

ament_tools/build_types/cmake.py
```diff
@@ -167,8 +167,9 @@ def _common_cmake_on_build(self, should_run_configure, context, prefix, extra_cm
             solution_file = solution_file_exists_at(
                 context.build_space, context.package_manifest.name)
             cmd = prefix + [MSBUILD_EXECUTABLE]
-            if '-j1' in context.make_flags:
+            if not '-j1' in context.make_flags:
```
mikaelarguedas: linter will certainly complain about using `not <variable> in` instead of `<variable> not in`

Thanks @codebot you just made my life much easier!

wjwwood: I fixed this in https://github.com/ament/ament_tools/pull/101. I guess the next line, `cmd += ['/m']` was added incorrectly before...

ament_tools/build_types/cmake.py
wjwwood: I changed it so that the BuildAction can take a custom env. This is slightly worse than putting the env in the `.bat` file that prefixes the command, because it's less obvious to the user that this flag has been switched on. Anyways, if we agree it's better, we can merge https://github.com/ament/ament_tools/pull/101 into this pr.

ament_tools/build_types/cmake.py
```diff
@@ -167,12 +167,27 @@ def _common_cmake_on_build(self, should_run_configure, context, prefix, extra_cm
...
+            ]
+            if msbuild_flags:
+                cmd += msbuild_flags
+                # If there is a parallelism flag in the msbuild_flags and it is not /m1,
+                # then turn on /MP for the compiler too.
+                if any(['/m' in x for x in msbuild_flags]) and '/m:1' not in msbuild_flags:
+                    env = dict(os.environ)
+                    if 'CL' in env:
+                        # make sure there isn't anything in there trying to prevent MP
+                        if not any(['/MP' in x for x in env['CL']]):
```
dirk-thomas: `in` is not a very defensive check. The substring could be anywhere. Maybe check for `startswith` instead.

codebot: addressed in next commit

ament_tools/build_types/cmake.py
```diff
@@ -167,12 +167,27 @@ def _common_cmake_on_build(self, should_run_configure, context, prefix, extra_cm
...
+            ]
+            if msbuild_flags:
+                cmd += msbuild_flags
+                # If there is a parallelism flag in the msbuild_flags and it is not /m1,
+                # then turn on /MP for the compiler too.
+                if any([x.startswith('/m') for x in msbuild_flags]) and '/m:1' not in msbuild_flags:
+                    env = dict(os.environ)
+                    if 'CL' in env:
+                        # make sure env['CL'] doesn't include an /MP already
+                        if not any([x.startswith('/MP') for x in env['CL']]):
```
dirk-thomas: Isn't `env['CL']` a string and not a list?

codebot: that's why I was using `in` before.

codebot: just added a call to `split()` to tokenize it first.

dirk-thomas:
What if the user sets `CL` to `/m1` / `/m:1`?

dirk-thomas:
With multiple processes `fastrtps` took 54s to build. In http://ci.ros2.org/job/ci_windows/1468/ which uses `--make-flags -j1` it took 160s.

+1 (after squashing).

wjwwood:
@codebot if you rebase this against master, since https://github.com/ament/ament_tools/pull/102 was merged, then the CI should start working.

wjwwood:
Lgtm

mikaelarguedas:
Can this be merged ? or do we need more testing/review?

wjwwood:
Bump, @codebot can this be merged?

wjwwood:
No, I was waiting on you. :smile: The person who opens it usually merges in my experience, but that's not set in stone.

PR #101, opened 2016-06-10
suggested changes to #100

wjwwood:
@codebot, for your consideration.

This passes the additional env variable only to the msbuild call, but has the downside that it is not in the prefix `.bat` file nor is it printed as an argument to the screen. A user trying to reproduce the behavior will probably not notice this subtle difference in how the tool is building the code and how they are.


wjwwood:
I haven't tested this though, so you might want to try it out before running with it.

dirk-thomas:
We could print the custom environment of the action to make it more visible.

dirk-thomas:
Since the user might have set `CL` in the environment this code should check for that (same as Linux / OS X checks for `MAKEFLAGS`).

wjwwood:
@codebot I think @dirk-thomas's comment still needs to be addressed.

dirk-thomas:
I don't think that is sufficient. The check needs to consider the content of `CL` in order to determine if appending `/MP` is intended or nor. Similar as for `MAKEFLAGS`.

PR #102, opened 2016-06-17
use python3.5 for travis testing

wjwwood:
Since we now use pydocstyle from python3.5 this is required.


wjwwood:
Also, apparently install `pydocstyle`.

wjwwood:
@dirk-thomas should we turn off pep257 tests? There are more than two hundred failures right now.

dirk-thomas:
I think it would be good to continue testing the code which already conforms. Maybe blacklist the third-party code which doesn't?

wjwwood:
Any idea why it is including those other ones? I thought this line should limit it to this package's source: https://github.com/ament/ament_tools/blob/master/test/test_pep257.py#L19

dirk-thomas:
It does so for me locally. So it might be related to the travis script.

wjwwood:
@dirk-thomas how did you determine that it is only checking those folders locally?

dirk-thomas:
Because it passed?

wjwwood:
I think this is a regression of ament/ament_lint#49, see: https://github.com/ament/ament_lint/pull/49/files#r67576895

wjwwood:
@dirk-thomas but that doesn't ensure that it only checks those directories, only that there are no other directories which need to be ignored.

dirk-thomas:
That's true. I thought the layout would be the same locally then in Travis.

If I place additional files in a new subfolder under `ament_tools` they are being linted by PEP257 which shouldn't be the case.

wjwwood:
Ok, I'll look into a fix later if you don't have time.

dirk-thomas:
Fixed in ament/ament_lint#57.

dirk-thomas:
@wjwwood FYI cca1c22f1639c48868789d63a07903a8607af920

wjwwood:
Thanks, I'll see if it works in f4c4c8f.

wjwwood:
lgtm, I'll revert f4c4c8f and +1 the fix.

dirk-thomas:
+1 (with squashing)

PR #104, opened 2016-07-01
apply --rerun-failed on Windows

dirk-thomas:
Addition to #98 which only added the flag for non Windows platforms.

Please review.


wjwwood:
lgtm

PR #105, opened 2016-07-01
add --parallel option

dirk-thomas:
While the interleaved output is not pretty to read the option achieves its goal to build packages in parallel and therefore speed up the build. The option is available for the `build` as well as the `test` verb. `--end` is still based on the sequential order and doesn't ignore packages which are before in the order but not recursive dependencies of the end package.

* http://ci.ros2.org/job/ci_linux/1604/
* http://ci.ros2.org/job/ci_osx/1217/
* http://ci.ros2.org/job/ci_windows/1588/

Please review.

### DIFF
--- a/ament_tools/setup_arguments.py
+++ b/ament_tools/setup_arguments.py
@@ -18,6 +18,9 @@ try:
     import setuptools
 except ImportError:
     pass
+from threading import Lock
+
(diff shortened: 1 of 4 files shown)


wjwwood:
Code changes look ok, but I didn't try it.

PR #106, opened 2016-09-20
error print when executing with non python3.5

Karsten1987:
addresses issue https://github.com/ament/ament_package/issues/42

Idea here is to give users feedback in case they trigger `ament.py` with a python version other than 3.5. If so, ament will exit with a syntax error. With this PR, the users gets some feedback about the wrong python version.

Totally up to discussion if the python version can be parsed differently or the location for that snippet should be moved.


dirk-thomas:
Users might run the script `scripts/ament.py` or install the package. So I think the check should be repeated in the `setup.py` file.

Karsten1987:
updated with all checks passed

PR #114, opened 2016-11-04
allow multiple package names for --only-packages

dirk-thomas:
* Extract checking the passed arguments for sane values and combinations.
* Enables multiple package names to be passed to `--only-packages`.
* Consolidates all the criteria which packages should be processed in a single attribute (`start_with`, `end_with`, `only_packages` are merged into `skip_packages`).
* Printing the topological order as well as the actual iteration over the packages then uses the consolidated result without the need to reimplement the logic.

### DIFF
--- a/ament_tools/verbs/build/cli.py
+++ b/ament_tools/verbs/build/cli.py
@@ -114,14 +114,14 @@ def prepare_arguments(parser, args):
         help='End with a particular package',
     )
     parser.add_argument(
-        '--only-package',
-        '--only',
-        help='Only process a particular package, implies --start-with <pkg> and --end-with <pkg>'
+        '--only-packages',
+        nargs='+', default=[],
+        help='Only process a particular set of packages'
     )
     parser.add_argument(
         '--skip-packages',
-        nargs='*',
-        help='List of packages to skip'
+        nargs='*', default=[],
+        help='Set of packages to skip'
     )
     parser.add_argument(
         '--parallel',
@@ -161,14 +161,20 @@ def main(opts, per_package_main=build_pkg_main):
         raise VerbExecutionError('Circular dependency within the following '
                                  'packages: %s' % circular_dependencies[0])
 
-    print_topological_order(opts, packages)
+    pkg_names = [p.name for _, p, _ in packages]
+    check_opts(opts, pkg_names)
+    consolidate_package_selection(opts, pkg_names)
+    print_topological_order(opts, pkg_names)
+
+    if set(pkg_names) <= set(opts.skip_packages):
+        print('All selected packages are being skipped. Nothing to do.',
+              file=sys.stderr)
+        return 0
 
     return iterate_packages(opts, packages, per_package_main)
 
 
-def print_topological_order(opts, packages):
-    package_names = [p.name for _, p, _ in packages]
-
+def check_opts(opts, package_names):
     if opts.start_with and opts.start_with not in package_names:
         sys.exit("Package '{0}' specified with --start-with was not found."
                  .format(opts.start_with))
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
Works fine locally, that's great! will save us quite some time
Question: Is there a rationale for being able to skip-package the package provides to `--end-with` but not one of the packages provided to `--only-packages`. Not saying there is a use case there but it's surprising that the behaviour is different while both packages are explicitely listed in commandline as "to be built"

dirk-thomas:
Listing a single pkg name in `--only-packages` as well as `--skip-packages` is contracting each other.

Using `start-with` / `--end-with` together with `--skip-packages` is no contradiction and can be reasonable (e.g. start with `D`, end with `H` but skip `F`).

mikaelarguedas:
Yeah I'm just surprised that I can run `--start-with A --end-with A --skip-package A` but I see your point

dirk-thomas:
When you consider a future option `--deps` it makes sense to ignore the starting package.

mikaelarguedas:
Fair point, we should just explain it somewhere in a future "ament tutorial" so that people are not confused by it.

dirk-thomas:
After talking about the various case more I updated the patch to only output a warning if all selected packages are also being skipped.

ament_tools/verbs/build/cli.py
```diff
@@ -114,13 +114,13 @@ def prepare_arguments(parser, args):
         help='End with a particular package',
     )
     parser.add_argument(
-        '--only-package',
-        '--only',
```
mikaelarguedas: Do we want to keep the `--only` alias for convenience ?

dirk-thomas: Any shorter but unique variation of arguments is always available. That's always why renaming it from `--only-package` to `--only-packages` is backward compatible.

ament_tools/verbs/build/cli.py
```diff
@@ -114,13 +114,13 @@ def prepare_arguments(parser, args):
...
-        help='Only process a particular package, implies --start-with <pkg> and --end-with <pkg>'
+        '--only-packages',
+        nargs='*', default=[],
+        help='Only process a particular set of packages'
```
dhood: `'List of packages to process'` would be more consistent with the --skip-packages help (and makes it clear it's not a flag)

ament_tools/verbs/build/cli.py
mikaelarguedas: I agree, but it won't be displayed in the help message anymore if it is removed here.

ament_tools/verbs/build/cli.py
dhood: I meant that the whole phrase makes it sounds like a boolean flag to me. but the array "type" in the help output will make it clearer

ament_tools/verbs/build/cli.py
```diff
@@ -177,77 +183,73 @@ def print_topological_order(opts, packages):
...
-        if skip_package == opts.only_package:
-            sys.exit("Cannot --skip-packages and --only-package the same package: '{0}'."
-                     .format(skip_package))
-    if nonexistent_skip_packages:
-            sys.exit('Packages [{0}] specified with --skip-packages were not found.'
-                     .format(', '.join(nonexistent_skip_packages)))
-
-    if opts.only_package:
+    nonexistent_skip_packages = set(opts.skip_packages) - set(package_names)
+    if set(opts.skip_packages) - set(package_names):
```
dirk-thomas: Updated.

ament_tools/verbs/build/cli.py
```diff
@@ -177,77 +183,73 @@ def print_topological_order(opts, packages):
...
                     sys.exit("The --end-with package '{0}' occurs topologically "
                              "before the --start-with package '{1}'"
                              .format(opts.end_with, opts.start_with))
                 break
 
-    print('# Topological order')
+
+def consolidate_package_selection(opts, package_names):
+    # after this function opts.skip_packages will contain the information from:
+    # start_package, end_package, only_packages
```
dirk-thomas: Sorry, documentation from a different tool which uses different argument names :wink: Updated.

dirk-thomas:
1. Since `--only-packages` is a greedy argument you need to insert `--` before passing other arguments behind it.

2. I changed the argument to `nargs='+'` to enforce passing at least one name.

3. Updated completion.

dhood:
I am still finding the greedy behaviour a bit strange when combined with `--skip-packages` and specifying the source path.

This works fine:
```
dhood@osrf-esteve:~/ros2_ws [ros2_ws]$ ament build --only-packages rclcpp rcl --skip-packages rcl -- ./src
```

But this does not:
```
dhood@osrf-esteve:~/ros2_ws [ros2_ws]$ ament build --only-packages rclcpp rcl -- --skip-packages rcl -- ./src
usage: ament build [-h] [-C DIRECTORY] [--build-space BUILD_SPACE]
                   [--install-space INSTALL_SPACE] [--build-tests]
                   [--make-flags [MAKE_FLAGS [MAKE_FLAGS ...]]] [--skip-build]
                   [--skip-install] [-s] [--isolated]
                   [--start-with START_WITH] [--end-with END_WITH]
                   [--only-packages ONLY_PACKAGES [ONLY_PACKAGES ...]]
                   [--skip-packages [SKIP_PACKAGES [SKIP_PACKAGES ...]]]
                   [--parallel] [--force-ament-cmake-configure]
                   [--ament-cmake-args [AMENT_CMAKE_ARGS [AMENT_CMAKE_ARGS ...]]]
                   [--force-cmake-configure]
                   [--cmake-args [CMAKE_ARGS [CMAKE_ARGS ...]]]
                   [--ctest-args [CTEST_ARGS [CTEST_ARGS ...]]]
                   [basepath]
ament build: error: argument basepath: Path '--skip-packages' does not exist
```

dirk-thomas:
The `--` tells argparse that the next argument is a positional argument. But the next one is `--skip-packages` so you must not precede it with `--` in that case.

dhood:
I see your point about `--` telling argparse that the optionals are done, and that what follows is the positional arguments. I agree that this is the expected behaviour in the context of argparse, and that I should not expect to use both `--only-packages` and `--skip-packages` with the delimiter.

This was unexpected for me because of the difference in how other optionals such as `--cmake-args` have to be passed.

To illustrate, here's a valid invocation for passing cmake args, only-packages, skip-packages and specifying the basepath:
```
~/ros2_ws$ ament build --cmake-args -DCMAKE_BUILD_TYPE=Debug -- --only-package rclcpp --skip-packages rcl rclcpp -- ./src
```

You'll see that, while more optionals come after --cmake-args, we have to pass the delimiter with it, since  [_we_ are parsing the delimiter](https://github.com/ament/ament_tools/blob/24a11a523cf900ce3880c62ed79127cc2177a4a4/ament_tools/build_types/cmake.py#L78) with [extract_argument_group](https://github.com/ament/ament_tools/blob/24a11a523cf900ce3880c62ed79127cc2177a4a4/ament_tools/helper.py#L124), not argparse.

This is why I figured you'd have to do the same for ending argument collection for `--only-package` in my previous comment.

I wouldn't have gotten into this case if `[basepath]` was listed first in the `usage` help output (as opposed to last). But I think people will be able to figure it out (I was just _trying_ to break the parsing :smile: )

PR #115, opened 2016-11-04
improve Python uninstall

dirk-thomas:
* Avoids to touch sibling folder with the same prefix (e.g. don't touch `foobar` when doing `foo`).
* Removes all recursive empty directories.

### DIFF
--- a/ament_tools/build_types/ament_python.py
+++ b/ament_tools/build_types/ament_python.py
@@ -369,14 +369,18 @@ class AmentPythonBuildType(BuildType):
         if os.path.exists(install_log):
             with open(install_log, 'r') as h:
                 lines = [l.rstrip() for l in h.readlines()]
-            directories = []
+            directories = set()
+            install_space = context.install_space + os.sep
             for line in lines:
                 if os.path.exists(line) and \
-                        line.startswith(context.install_space):
+                        line.startswith(install_space):
                     if not os.path.isdir(line):
                         os.remove(line)
-                    else:
-                        directories.append(line)
+                        while True:
+                            line = os.path.dirname(line)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Waiting for review...

ament_tools/build_types/ament_python.py
```diff
@@ -369,14 +369,18 @@ def _undo_install(self, context):
         if os.path.exists(install_log):
             with open(install_log, 'r') as h:
                 lines = [l.rstrip() for l in h.readlines()]
-            directories = []
+            directories = set()
+            install_space = context.install_space + os.sep
             for line in lines:
                 if os.path.exists(line) and \
-                        line.startswith(context.install_space):
+                        line.startswith(install_space):
```
dhood: Have you thought about what will happen when `install_space` has quotes, spaces, and slashes? (It might work fine still - I am investigating it myself, but you may have thought about it already)

dirk-thomas: I don't see a problem with that. Important (and missing before) is to include the trialing `/` when doing the `startswith`.

dirk-thomas: I don't know if there are cases on Windows where that is the case. Since the code was doing the same before (checking with `startswith`) I don't plan to look into that in the context of this PR.

dhood: ok no problem

dhood:
This looks good to me. Note that I couldn't test this PR rebased on master because the `uninstall` verb has broken since this PR was opened - ticketed in https://github.com/ament/ament_tools/issues/122

PR #125, opened 2016-11-24
report skipped tests

mikaelarguedas:
connects to https://github.com/ros2/ros2/issues/290


ament_tools/verbs/test_results/cli.py
```diff
@@ -160,12 +162,12 @@ def print_summary(results, show_stable=False, show_unstable=True):
...
-        if show_unstable and (num_errors or num_failures):
-            print('%s: %d tests, %d errors, %d failures' %
-                  (name, num_tests, num_errors, num_failures))
-    sum_tests, sum_errors, sum_failures = aggregate_results(results, callback)
-    print('Summary: %d tests, %d errors, %d failures' %
-          (sum_tests, sum_errors, sum_failures))
+            print('%s: %d tests, %d skipped' % (name, num_tests, num_skipped))
+        if show_unstable and (num_errors or num_failures or num_skipped):
+            print('%s: %d tests, %d skipped, %d errors, %d failures' %
+                  (name, num_tests, num_skipped, num_errors, num_failures))
```
dirk-thomas: With the current logic it seems that both pint statements (line 167 as well as 169) could happen since the second part of the boolean condition is not the exact inverse anymore?

mikaelarguedas: indeed, I guess that given that skipped tests don't impact the stability of the build they should just not be part of the boolean condition at all and displayed in both cases. fixed in https://github.com/ament/ament_tools/pull/125/commits/1d46b77536d8f4066eb9d919e1524cb8e62ad293

dirk-thomas: Sounds good to me.

Issue #88, opened 2016-04-05
[cmake] use `cmake --build` rather than calling `make` or `msbuild` directly


wjwwood:
There are a few complications with this, such as passing "make arguments" through `cmake --build` (when using `make`), but I think it will make our tools more flexible.

ahundt:
Thanks!

jpgr87:
`cmake --build` allows you to [pass flags to the native tool](https://cmake.org/cmake/help/v3.9/manual/cmake.1.html#build-tool-mode) by using a `--` on the command line.  Is that what you mean for passing "make arguments"?

wjwwood:
@jpgr87 yes that's what I was talking about. I meant it just means that only part of the build command is portable, which the point of this was, i.e. don't assume make or ninja or msbuild. But since the parallelization flags are not portable you still need to know which you are using so you know which flags to pass cmake. So "just using `--build`" doesn't remove any need to parse and know which cmake generator you used.

ahundt:
@wjwwood am I correct that `cmake --build` would assist with building when using generators that aren't via a hard coded command? If you don't care about the parallelization flags that's useful.

I don't really have context for the prior conversation, so apologies if my comment is nonsensical. :-)

dirk-thomas:
@ahundt We will hopefully get back to switching to a universal build tool and will not continue to support `ament_tools`. I just wanted to mentioned this before your are considering to spend significant effort on improving it. My current prototype of the new build tool already uses `cmake --build` and supports different generators.

ahundt:
Awesome! FYI if you're reconsidering options for a universal build tool, homebrew has improved dramatically over the last year and a half and now supports Linux, has more a more OS agnostic backend, and has solved the major issues with multiple versions of the same project. Perhaps the effort for adding windows would be low enough bar and then homebrew could meet your goals?

- https://github.com/Homebrew/brew/releases/
- https://github.com/Linuxbrew/brew/issues/1

No worries if it still does not make sense to use homebrew. :-)

wjwwood:
> If you don't care about the parallelization flags that's useful.

But we do 😛.

dirk-thomas:
Closing since the repo is archived.

PR #127, opened 2016-12-08
Use Xcode as system generator and builder on OSX

esteve:
This PR replaces `make` with `xcodebuild` as the builder for OSX, this among other things allows for ROS2 to be compiled for iOS and the iOS Simulator.

I kept the name of the builder flags option (`make-flags`), but I think that if this gets merged it may make sense to rename it to something else (e.g. `builder-flags`), since Linux would be the only platform to actually use `make`


esteve:
@wjwwood sorry for pinging you specifically, but given that you're the one who uses OSX to work on ROS2, I thought you'd the most adequate for reviewing this. However, if anyone else on the @ros2/team feels like having a look at the changes, please feel free to do so.

I believe the changes in this PR can be merged because they:
- do not break existing code and all the existing usecases are supported (e.g. parallel building). IIRC last time I ran this on local, everything worked fine, hopefully nothing has changed.
- add support for using ROS 2.0 as an XCode project (e.g. debugging from the IDE)
- add support for building ROS 2.0 on iOS. The iOS SDK does not support Make AFAIK.

wjwwood:
@esteve the changes look ok to me, but I would like to preserve the capability to build with make on macOS, something like a `--use-xcode` option for `ament build`. I'll try to take a stab at implementing this as an option as soon as I can.

wjwwood:
Just for some of my thinking behind that desire... Xcode is not currently a requirement for building ROS 2 because you can just install "Command Line Tools" and avoid Xcode, which requires an apple account (I think that's still the case anyways). Also, I like to go into my build folders and run `make ...` 😄.

But still, the ability to run in Xcode is pretty nice. Luckily I don't think it will be difficult to make this switch optional.

esteve:
@wjwwood probably bad phrasing on my part in my previous comment, but Xcode is not really a requirement for this. `xcodebuild`, which replaces `make` on this PR, is part of the "Command Line Tools" as well, so when you install Homebrew (which is a requirement for ROS 2.0), you already have it. So you can still build from the command line using either `xcodebuild` or `cmake --build`. I don't know if I still have permissions to trigger a CI build, but this change should still work on the buildfarm, which does not have Xcode installed.

Another option, if you want to keep the Make-based system is to check what generator was used (CMake's `-G` flag), and let `ament` choose `make` or `xcodebuild` accordingly.

esteve:
@wjwwood what I mean is that, analogously to this code:

https://github.com/esteve/ament_tools/blob/xcode/ament_tools/build_types/cmake.py#L410-L435

something similar could be done in the build phase:

```
        elif IS_MACOSX:
            # Check CMake was invoked to generate a Xcode or Make project
            # The Xcode CMake generator will produce a file named
            # ALL_BUILD_cmakeRulesBuildPhase.makeRelease if the Xcode generator
            # was used, whether the build type is Release, Debug or anything else
            all_build_cmake_file_path = os.path.join(
                context.build_space, 'CMakeScripts', 'ALL_BUILD_cmakeRulesBuildPhase.makeRelease')
            all_build_cmake_file = os.path.isfile(all_build_cmake_file_path)
            if all_build_cmake_file:
                # This is an Xcode project, use xcodebuild to compile
                ...
                yield BuildAction(prefix + [XCODEBUILD_EXECUTABLE] + xcodebuild_args)
            else:
                # Assume that CMake generated a Make-based project
                ...
                yield BuildAction(prefix + [MAKE_EXECUTABLE] + context.make_flags)
```

and default to the Make generator. IMHO `xcodebuild` is better than Make (anything is better than Make by default :smile: ) and both are abstracted by `cmake --build`, but it's true that people are used to Make and it's hard to change habits, so if it causes less friction, I'm fine to set Make as the default type for the CMake generator.

What do you think? I can write that logic, but it'd be good to run a CI build with Xcode as the generator as well.

wjwwood:
I don't actually mind what the default is on macOS (as long as it comes with CLT and doesn't introduce regressions), but I do think there should be a way to use `make` if you want, whether or not that's the default.

esteve:
@wjwwood I've updated the PR to support both Make and Xcode, I've left Xcode as the default, but it can easily be changed to Make here:

https://github.com/esteve/ament_tools/blob/xcode/ament_tools/build_types/cmake.py#L171-L172

to override the default, the user can pass the `-G` flag to `--cmake-args` or `--ament-cmake-args`.

However, I had always tested with PR without `symlink-install` and I just noticed that it won't work because the Xcode generator is a multi-config generator (https://cmake.org/cmake/help/v3.5/variable/CMAKE_BUILD_TYPE.html#variable:CMAKE_BUILD_TYPE and https://cmake.org/cmake/help/v3.5/variable/CMAKE_CONFIGURATION_TYPES.html), and thus creates multiple conflicting targets if `ament` is asked to create symbolic links for the install step. I ended up adding the following to `ament_cmake` https://github.com/esteve/ament_cmake/commit/9c27a5201b50fe9f759172bfcc1ab1e219dfb0d3, so I don't know if it's just better to keep Make as the default generator/builder for CMake.

dirk-thomas:
With the downside of not being able to use the symlink installs (not sure if that could be fixed, I don't have a platform with multi-config to test this on) I would propose to keep `make` the default.

> to override the default, the user can pass the -G flag to --cmake-args or --ament-cmake-args.

I would suggest adding a option `--use-xcode` as it is done in ROS 1 for other generators already.

Please see #144 for a different PR adding support for Ninja which will likely interfere with this.

esteve:
@dirk-thomas I agree, it makes sense. I've made the Make generator the default again (and more explicit see https://github.com/esteve/ament_tools/blob/xcode/ament_tools/build_types/cmake.py#L181) and added a `--use-xcode` option to `ament` which is only available when run on macOS.

Since it's not the default, I also added `-DCMAKE_XCODE_GENERATE_SCHEME=ON`, which will be available in CMake 3.8.0. The new scheme generator for CMake 3.8 adds better support for `xcodebuild test`, for example. This way, the default behavior will still be compatible with the minimum requirements (i.e. CMake 3.5) and only the optional `--use-xcode` flag requires a greater version of CMake.

As for the issue with multi-config generators and symlink installs, should I submit a PR with the change in https://github.com/esteve/ament_cmake/commit/9c27a5201b50fe9f759172bfcc1ab1e219dfb0d3 ?

dirk-thomas:
> As for the issue with multi-config generators and symlink installs, should I submit a PR with the change in esteve/ament_cmake@9c27a52 ?

It would be good to separate that change. I don't see why the symlinks install should need to be disabled. It is supposed to replace the normal install rules transparently - and those obviously work for multi configuration generators. So it looks like the symlinks install stuff needs to be modified somehow to work in that scenario.

esteve:
> and those obviously work for multi configuration generators.

@dirk-thomas this issue hadn't show up yet because the only multi config generator that's in use in ROS 2.0 is Visual Studio's and symlinks are not supported in Windows. However, macOS supports symlinks but the Xcode generator fails with the code here:

https://github.com/ament/ament_cmake/blob/master/ament_cmake_core/cmake/symlink_install/ament_cmake_symlink_install_targets.cmake#L96-L98

invoking ament with `src/ament/ament_tools/scripts/ament.py build --use-xcode --build-tests --symlink-install --isolated --build-space build_isolated_xcode --install-space install_isolated_xcode` produces the following error:

```
+++ Building 'class_loader'
==> '. /Users/esteve/ros2_ws/build_isolated_xcode/class_loader/cmake__build.sh && /usr/local/bin/cmake /Users/esteve/ros2_ws/src/ros/class_loader -DBUILD_TESTING=1 -DAMENT_CMAKE_SYMLINK_INSTALL=1 -DCMAKE_INSTALL_PREFIX=/Users/esteve/ros2_ws/install_isolated_xcode/class_loader -G Xcode -DCMAKE_XCODE_GENERATE_SCHEME=ON' in '/Users/esteve/ros2_ws/build_isolated_xcode/class_loader'
-- Found ament_cmake: 0.0.0 (/Users/esteve/ros2_ws/install_isolated_xcode/ament_cmake/share/ament_cmake/cmake)
-- Using PYTHON_EXECUTABLE: /usr/local/bin/python3
-- Override CMake install command with custom implementation using symlinks instead of copying resources
-- Found poco_vendor: 1.0.0 (/Users/esteve/ros2_ws/install_isolated_xcode/poco_vendor/share/poco_vendor/cmake)
-- Searching for Poco library...
--   Found Poco!
-- components found: Foundation.
-- Configuring done
CMake Error in CMakeLists.txt:
  Evaluation file to be written multiple times for different configurations
  or languages with different content:

    /Users/esteve/ros2_ws/build_isolated_xcode/class_loader/ament_cmake_symlink_install_targets_0.cmake


CMake Error in CMakeLists.txt:
  Evaluation file to be written multiple times for different configurations
  or languages with different content:

    /Users/esteve/ros2_ws/build_isolated_xcode/class_loader/ament_cmake_symlink_install_targets_0.cmake


CMake Error in CMakeLists.txt:
  Evaluation file to be written multiple times for different configurations
  or languages with different content:

    /Users/esteve/ros2_ws/build_isolated_xcode/class_loader/ament_cmake_symlink_install_targets_0.cmake


CMake Error in CMakeLists.txt:
  Evaluation file to be written multiple times for different configurations
  or languages with different content:

    /Users/esteve/ros2_ws/build_isolated_xcode/class_loader/ament_cmake_symlink_install_targets_0.cmake


CMake Error in CMakeLists.txt:
  Evaluation file to be written multiple times for different configurations
  or languages with different content:

    /Users/esteve/ros2_ws/build_isolated_xcode/class_loader/ament_cmake_symlink_install_targets_0.cmake


CMake Error in CMakeLists.txt:
  Evaluation file to be written multiple times for different configurations
  or languages with different content:

    /Users/esteve/ros2_ws/build_isolated_xcode/class_loader/ament_cmake_symlink_install_targets_0.cmake


-- Generating done
-- Build files have been written to: /Users/esteve/ros2_ws/build_isolated_xcode/class_loader

<== Command '. /Users/esteve/ros2_ws/build_isolated_xcode/class_loader/cmake__build.sh && /usr/local/bin/cmake /Users/esteve/ros2_ws/src/ros/class_loader -DBUILD_TESTING=1 -DAMENT_CMAKE_SYMLINK_INSTALL=1 -DCMAKE_INSTALL_PREFIX=/Users/esteve/ros2_ws/install_isolated_xcode/class_loader -G Xcode -DCMAKE_XCODE_GENERATE_SCHEME=ON' failed in '/Users/esteve/ros2_ws/build_isolated_xcode/class_loader' with exit code '1'
<== Command '. /Users/esteve/ros2_ws/build_isolated_xcode/class_loader/cmake__build.sh && /usr/local/bin/cmake /Users/esteve/ros2_ws/src/ros/class_loader -DBUILD_TESTING=1 -DAMENT_CMAKE_SYMLINK_INSTALL=1 -DCMAKE_INSTALL_PREFIX=/Users/esteve/ros2_ws/install_isolated_xcode/class_loader -G Xcode -DCMAKE_XCODE_GENERATE_SCHEME=ON' failed in '/Users/esteve/ros2_ws/build_isolated_xcode/class_loader' with exit code '1'
```

ament_tools/build_types/cmake.py
```diff
@@ -188,13 +216,62 @@ def _common_cmake_on_build(self, should_run_configure, context, prefix, extra_cm
                         env['CL'] = '/MP'
             cmd += [
                 '/p:Configuration=%s' %
-                self._get_visual_studio_configuration(context), solution_file]
+                self._get_configuration_from_cmake(context), solution_file]
             yield BuildAction(cmd, env=env)
+        elif IS_MACOSX:
+            if self._using_xcode_generator(context):
+                if XCODEBUILD_EXECUTABLE is None:
+                    raise VerbExecutionError("Could not find 'xcodebuild' executable")
```
dirk-thomas: Does this prevent building with `Makefiles` when `xcodebuild` is not available?

ament_tools/build_types/cmake.py
```diff
@@ -188,13 +216,62 @@ def _common_cmake_on_build(self, should_run_configure, context, prefix, extra_cm
...
+                cmd += ['build']
+                xcodebuild_flags = []
+                for flag in context.make_flags:
+                    if not flag.startswith('-l'):
+                        if flag.startswith('-j'):
+                            xcodebuild_flags.append(
+                                flag.replace(
+                                    '-j', '-IDEBuildOperationMaxNumberOfConcurrentCompileTasks='))
+                        else:
+                            xcodebuild_flags.append(flag)
```
dirk-thomas: The logic is a bit difficult to follow. Shouldn't this be the equivalent of:

```
if flag.startswith('-j'):
    xcodebuild_flags.append(flag.replace(
        '-j', '-IDEBuildOperationMaxNumberOfConcurrentCompileTasks='))
elif not flag.startswith('-l'):
    xcodebuild_flags.append(flag)
```

ament_tools/build_types/cmake.py
```diff
@@ -368,37 +443,99 @@ def _common_cmake_on_install(self, context):
...
+                            if flag.startswith('-j'):
+                                xcodebuild_flags.append(
+                                    flag.replace(
+                                        '-j',
+                                        '-IDEBuildOperationMaxNumberOfConcurrentCompileTasks='))
+                            else:
+                                xcodebuild_flags.append(flag)
+                    xcodebuild_flags += ['-configuration']
+                    xcodebuild_flags += [self._get_configuration_from_cmake(context)]
+                    cmd.extend(xcodebuild_flags)
```
dirk-thomas: Almost the logic has been added above. This should be deduplicated.

Another time below.

ament_tools/build_types/cmake.py
wjwwood: Just based on the code, I think this only happens when you've specified that you want to use the xcode generator.

ament_tools/build_types/cmake.py
```diff
@@ -149,20 +173,24 @@ def _common_cmake_on_build(self, should_run_configure, context, prefix, extra_cm
                 if vsv not in supported_vsv:
                     raise VerbExecutionError('Unknown / unsupported VS version: ' + vsv)
                 cmake_args += ['-G', supported_vsv[vsv]]
+            elif IS_MACOSX:
+                if context.use_xcode or self._using_xcode_generator(context):
+                    cmake_args += ['-G', 'Xcode']
+                    cmake_args += ['-DCMAKE_XCODE_GENERATE_SCHEME=ON']
```
wjwwood: This one assumes CMake 3.8.0 right? Could we conditionally add that with that version only? Or is there no point in running without it? My impression was that it just improved testing with Xcode?

ament_tools/build_types/cmake.py
wjwwood: 👍, the altered logic is easier to read I think.

ament_tools/build_types/cmake.py
wjwwood: I agree that some of this logic could be consolidated with the code above. Especially the handling of the `context.make_flags`.

esteve: Agreed, I've added @dirk-thomas 's code and moved it to a separate function in https://github.com/ament/ament_tools/pull/127/commits/9ace6e8203c2e333ce37b618935073e15a578bad

Thanks!

ament_tools/build_types/cmake.py
esteve: I've removed the `CMAKE_XCODE_GENERATE_SCHEME` flag and used the `RUN_TESTS` target instead for `ament test` as a workaround in https://github.com/ament/ament_tools/pull/127/commits/eaf9430d79f3f1af4db1fd98c23cbffe53cd9180 This approach is compatible with CMake 3.5, so no need to use anything experimental from 3.8

esteve:
@dirk-thomas @wjwwood I've addressed your feedback, let me know if there's anything else that needs to be fixed/improved. Thanks!

esteve:
@dirk-thomas @wjwwood I just realized that the duplicated code was not really needed in the `install` and `uninstall` verbs, so I moved the code back to `_common_cmake_on_build`

dirk-thomas:
Not directly related to this patch but a lot of this custom logic would be unnecessary if we would implement #88. But since `ament_tools` is likely not being developed much further (see http://design.ros2.org/articles/build_tool.html) I don't think we will spend time to migrate to `cmake --build`. Since this patch addresses the short term desire to use XCode that's fine with me.

I will leave it up to @wjwwood to try it on OS X and comment / merge this patch.

wjwwood:
I tested the `--use-xcode` option locally and it works!

I also ran CI with the default behavior to see if breaks anything, but it works (after 0371b58):

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2414/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1864/) (flaky tests make it unstable)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2534/) (flaky tests, compared to [similar nightly failures](http://ci.ros2.org/view/nightly/job/nightly_win_rep/610/testReport/(root)/test_showimage_cam2image__rmw_connext_cpp_Release/test_reliable_qos/))

But of course just as I get it tested, it's all broken now that @dirk-thomas pushed https://github.com/ament/ament_tools/pull/144 through.

I'll have a look at the conflict resolution...

wjwwood:
Passing CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2416/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1870/) (with some flaky tests)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2538/) (also with some flaky tests)

wjwwood:
Thanks again @esteve!

esteve:
Yay! Thanks @wjwwood for testing this and for fixing the conflicts.

esteve:
And thanks @dirk-thomas for the review.

PR #130, opened 2017-01-05
Don't add cover-inclusive option

dhood:
If an ament_python package has a python file that is only conditionally supposed to be used, the `cover-inclusive` option for nose can cause coverage tests to fail.

An example is https://github.com/ament/ament_lint/pull/63/commits/bf07f5b99ebcab3f0fa8abe7a807458ca327afb7 where there is a module that is only supposed to be used if a particular version of a package is installed. If another version is installed when the tests are run, coverage fails.

Am I right in thinking that this flag is just a "nice to have", but is adding more requirements than necessary on ament_python packages? Or should all ament_python packages be structured so that the coverage works with `--cover-inclusive`?

To check that we are not relying on this for finding test files: the test count for with ([Build Status](http://ci.ros2.org/job/ci_linux/2190/)) and without ([Build Status](http://ci.ros2.org/job/ci_linux/2190/)) this change are the same. (Some tests fail because I didn't put `--retest-until-pass` because it affects the test count)


dirk-thomas:
We added the option because we thought this would be benefitiial since otherwise the coverage report might be incomplete. Without the coverage job running it is difficult to make a decision on this.

mikaelarguedas:
Am I right to suppose that, once we move to a structure of separated code and test, --e.g. human readable code in some repos and testing boilerplate somewhere else-- the coverage of python packages will be almost inexistant if we get rid of this flag ? I'm not familiar with `cover-inclusive` hence the question.

dirk-thomas:
I am not sure what you mean with:

>  once we move to a structure of separated code and test, --e.g. human readable code in some repos and testing boilerplate somewhere else

E.g. for a Python package like `ament_package` the tests will always remain inside the package. Assume the package has two files `a.py` and `b.py` (which don't import each other. If we wouldn't use `--cover-inclusive` and write a test which only imports `a.py` it can reach 100% code coverage without ever testing a single line in `b.py`.

dhood:
linters for ament_python packages have to be added manually (e.g. https://github.com/ament/ament_tools/blob/5c5c005566212d7f6fd7ed739ee409c698dfdfe7/test/test_copyright.py)

Instead of automatically adding the coverage test for all ament_python packages, should we require them to be explicitly added by users, as is the case for linters?

dirk-thomas:
How do you suggest to make the coverage option a per package choice?

mikaelarguedas:
If there is a way to implement the logic to run coverage only on a set of
packages I think I would prefer to explicitly blacklist packages that are
breaking the coverage job. In other words run coverage by default so that
skipping coverage is something that needs an extra step for the developer
rather than just forgetting to add coverage to the package.

On Jan 10, 2017 08:02, "Dirk Thomas" <notifications@github.com> wrote:

How do you suggest to make the coverage option a per package choice?

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/ament/ament_tools/pull/130#issuecomment-271615442>, or mute
the thread

dirk-thomas:
We can certainly blacklist problematic package for now in the coverage job. But ultimately we do want coverage statistics for those too. Imo the easiest way to achieve that would be to make them work with the `--coverage-inclusive` option.

dhood:
> How do you suggest to make the coverage option a per package choice?
The same way it's done for linters. people would have to add, e.g., a `test_coverage` file that invokes a coverage test, like they add a `test_pep8` script that invokes pep8. The `COVERAGE_FILE` environment variable is set by ament_tools, so the file should still end up in the build directory as before, but each package would have control over which options are passed.

> In other words run coverage by default so that
> skipping coverage is something that needs an extra step for the developer
> rather than just forgetting to add coverage to the package.
I agree this is good for when it's our packages. For users, though, I think that making all ament_python packages run coverage with no way to disable it is not the final goal. For users, it should be opt-in somehow.

> Imo the easiest way to achieve that would be to make them work with the --coverage-inclusive option.
Yes, absolutely. but this is the second time that I've run into surprises because coverage tests are being run when I didn't expect them to be, so I thought it would be good to open the discussion.

dirk-thomas:
> > How do you suggest to make the coverage option a per package choice?

> The same way it's done for linters. people would have to add, e.g., a `test_coverage` file that invokes a coverage test, like they add a `test_pep8` script that invokes pep8. The `COVERAGE_FILE` environment variable is set by ament_tools, so the file should still end up in the build directory as before, but each package would have control over which options are passed.

I don't think I understand the approach yet. The coverage options are being invoked on `nose` which then calls all the test files. I don't see how a single test file could influence that. Maybe it is best if you could propose a PR to see how this would work?

dhood:
https://github.com/ament/ament_lint/pull/63 has been resolved without the need for this PR so I will close this and we can continue discussion at a future date if appropriate

Issue #132, opened 2017-02-18
Sourcing overlay workspace's setup file isn't enough for python packages


dhood:
If you follow this tutorial https://github.com/ros2/ros2/wiki/Ament-Tutorial the part where it says 

>If you are returning with a new terminal to your development and want to pick up developing on your overlay you can simply source ~/overlay_ws/setup.bash which will source all parent workspaces environments automatically.
 
works fine for ament_cmake packages but doesn't seem to apply to ament_python packages.

I understand this to be related to https://github.com/ament/ament_tools/issues/109

(The following outputs are from a while ago, but https://github.com/ros2/ros2/issues/312, which was opened today, is the same issue)

```
dhood@osrf-esteve:~/overlay_ws [overlay_ws]$ . install/setup.bash 
Warning: OSPL_URI was already set to [[file:///usr/etc/opensplice/config/ospl.xml]]. This will not override it to the default [[file:///home/dhood/ros2_ws1/install_isolated/opensplice_cmake_module/share/opensplice_cmake_module/config/ros_ospl.xml]]. Please make sure this is the config that you want.
dhood@osrf-esteve:~/overlay_ws [overlay_ws]$ talker_py 
Traceback (most recent call last):
File "/home/dhood/overlay_ws/install/bin/talker_py", line 9, in 
load_entry_point('rclpy-examples==0.0.0', 'console_scripts', 'talker_py')()
File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 542, in load_entry_point
return get_distribution(dist).load_entry_point(group, name)
File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2569, in load_entry_point
return ep.load()
File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2229, in load
return self.resolve()
File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2235, in resolve
module = __import__(self.module_name, fromlist=['__name__'], level=0)
File "/home/dhood/overlay_ws/install/lib/python3.5/site-packages/topics/talker_py.py", line 18, in 
import rclpy
File "/home/dhood/ros2_ws1/install_isolated/rclpy/lib/python3.5/site-packages/rclpy/__init__.py", line 23, in 
from rclpy.impl import rmw_implementation_tools
[...]

PR #137, opened 2017-03-01
add AMENT_PREFIX_PATH env hook for ament_python packages

wjwwood:
Fixes https://github.com/ament/ament_tools/issues/132

@dirk-thomas I think this was just an oversight on our part, or at least I can't think of, or remember, a reason that `ament_python` packages aren't placed on the `AMENT_PREFIX_PATH`.


dirk-thomas:
For CMake the `ament_cmake` build type takes care of this within CMake. For plain CMake build types this was added in #93.

The idea for Python was similar. An `ament_python` build type should take of this within the `setup.py` file on its own (e.g. by using a custom `setup` function provided by some ament package). Only for a plain `python` build type the tool should need to perform this.

 The reason why the package itself should take care of it when possible is that the user could skip using the tool and use the native commands to build a package and it should stil work. Currently there is no `python` build type but maybe this is a good reason to follow up on this and create two distinct build types as well as the custom setup function.

wjwwood:
I agree that there should be a `python` and an `ament_python` build type. However, that's a separate issue from the fact that the current strategy (which deploys these files through the tool rather than as part of the `setup.py`) is missing an env hook for the APP.

It seems that you agree that `ament_python` packages should contribute to the APP, is that right?

I started CI for this:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2272/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1748/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2272/)

dirk-thomas:
> It seems that you agree that ament_python packages should contribute to the APP, is that right?

Yes, as a temporary work around.

PR #142, opened 2017-03-22
Ament set environm var when running.

barbalberto:
See issue https://github.com/ament/ament_cmake/issues/56

When ament is running, the env var `BUILT_WITH_AMENT_BUILD` is set to `true`, so that cmake or other builder can check whether they were invokated by ament or not.
This is useful for example when migrating from a different build system (catkin or plain cmake) to ament because the package is aware of what's going on and can do adjustments to cope with different situations. In my case the installation folder for config files changes; in this way I can do it correctly.

The env var is visible only to threads/processes started by ament itself. When ament terminates, the shell it reeturns to does not have the env var of course.

### DIFF
--- a/ament_tools/commands/ament.py
+++ b/ament_tools/commands/ament.py
@@ -14,6 +14,7 @@
 
 import argparse
 import sys
+import os
 
 from ament_tools.verbs import VerbExecutionError
 
@@ -29,6 +30,8 @@ VERBS_ENTRY_POINT = '{0}.verbs'.format(COMMAND_NAME)
 def main(sysargs=None):
     # Assign sysargs if not set
     sysargs = sys.argv[1:] if sysargs is None else sysargs
+
+    os.environ['AMENT_TOOLS'] = '1'
 
     # Create a top level parser
     parser = argparse.ArgumentParser(


dirk-thomas:
The patch sets the variable globally in the command. The name though implies that it is only set when running the `build` verb. Either the location where the environment variable is being set or name of the environment variable has to change to match.

barbalberto:
Thanks @dirk-thomas for the feedback.
I see what you mean and I'm fine to change the name or the scope (or both).

I set it globally because I think it'll be useful at least for the `build`, `build_pkg`, `uninstall`, `uninstall_pkg` verbs and probably also in `test` and `test_pkg`.
In case new verbs will be introduced they will have it for free, so we don't have to remember each time to add it (or not).
In general I think an addictional env var should not harm (although env pollution is not good)

I took the name `BUILT_WITH_AMENT_BUILD` as emerged in the discussion linked above, any name is good to me.

Personally, I'd choose a general name and a global scope, to cope with all possible use cases we don't image yet.
What do you think?

dirk-thomas:
How about naming the environment variable `AMENT_TOOLS` then?

barbalberto:
Ok, updated using `AMENT_TOOLS` as var name.

ament_tools/commands/ament.py
```diff
@@ -30,6 +31,8 @@ def main(sysargs=None):
     # Assign sysargs if not set
     sysargs = sys.argv[1:] if sysargs is None else sysargs
 
+    os.environ["AMENT_TOOOLS"] = "true"
```
dirk-thomas: Spelling: `AMENT_TOOLS`

Nitpick: when updating the line please also use single quotes instead of double quotes.

barbalberto: Ops, sorry. Updated

ament_tools/commands/ament.py
```diff
@@ -30,6 +31,8 @@ def main(sysargs=None):
     # Assign sysargs if not set
     sysargs = sys.argv[1:] if sysargs is None else sysargs
 
+    os.environ['AMENT_TOOLS'] = 'true'
```
dirk-thomas: Maybe set it to `1` to not give the wrong expectation that setting it to `false` be the opposite?

barbalberto: Done!

dirk-thomas:
Thanks for the patch and for iterating on it.

mikaelarguedas:
does this fix https://github.com/ament/ament_cmake/issues/56 or is it only implementing a subset of it?

barbalberto:
@mikaelarguedas  This env var can be seen by all ament verbs. I'm using it inside a plain CMake project

PR #147, opened 2017-05-10
add option to pass custom Python interpreter

dirk-thomas:
Connect to ros2/ci#72

### DIFF
--- a/ament_tools/verbs/build_pkg/cli.py
+++ b/ament_tools/verbs/build_pkg/cli.py
@@ -185,6 +185,13 @@ def add_arguments(parser):
         default=False,
         help='Use symlinks instead of copying files wherever possible',
     )
+    parser.add_argument(
+        '--python-interpreter',
+        default=sys.executable,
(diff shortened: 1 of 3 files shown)


ament_tools/verbs/build_pkg/cli.py
```diff
@@ -185,6 +185,13 @@ def add_arguments(parser):
         default=False,
         help='Use symlinks instead of copying files wherever possible',
     )
+    parser.add_argument(
+        '--python-interpreter',
+        default=sys.executable,
+        help='Use a different Python interpreter for invoking subprocesses'
+             " (default '%s')" % sys.executable,
```
wjwwood: nitpick: using different quote types on strings that are implicitly concatenated looks weird, I'd recommend changing the first lint to use `"` too.

dirk-thomas: Sure. Done.

wjwwood:
Looks like CI is failing due to an unused import.

dirk-thomas:
Fixed CI.

Issue #109, opened 2016-10-17
need to source build dependencies for ament-python build type


mikaelarguedas:

dirk-thomas:
This seems to be already the case: see https://github.com/ament/ament_tools/blob/711fa60ffaa711d5d4f26ea5bcef6e794bfd9cae/ament_tools/build_types/ament_python.py#L152

Can you please provide a reproducible example.

dirk-thomas:
Might it be related to the introspection of the setup file before the actual build is happening? Triggered by this line https://github.com/ament/ament_tools/blob/711fa60ffaa711d5d4f26ea5bcef6e794bfd9cae/ament_tools/build_types/ament_python.py#L470

mikaelarguedas:
TBH that was a long time ago, I don't really remember. I think that I needed to source manually rclpy otherwise I could not build the executables of https://github.com/ros2/examples/issues/127 (because setup.py was calling some rclpy functions)

clalancette:
@mikaelarguedas Since the upstream PR was merged, can this be closed?

wjwwood:
@dirk-thomas when you get a chance, can this be merged? (@mikaelarguedas said to ask you)

PR #149, opened 2017-06-19
move common prefix logic to base class

dirk-thomas:
Refactoring to deduplicate logic. No functional change.

**Update:** see comments below for functional changes in the third commit.

Connect to ros2/ros2#364.

### DIFF
--- a/ament_tools/build_type.py
+++ b/ament_tools/build_type.py
@@ -210,42 +210,52 @@ class BuildType(object):
     ):
         if additional_dependencies is None:
             additional_dependencies = []
-        lines = []
+        return get_command_prefix(
+            '%s__%s' % (build_type, name),
+            context.build_space,
+            dependencies=context.build_dependencies + additional_dependencies,
+            additional_lines=additional_lines)
+
+
+def get_command_prefix(
+    generated_filename, build_space, dependencies, additional_lines=None
+):
+    lines = []
+    if not IS_WINDOWS:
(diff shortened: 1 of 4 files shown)


dirk-thomas:
The third commit addresses #109.

Since this introduces a significant change I am asking for a re-review.

wjwwood:
Doh, I didn't see ded89e7. Let me look at it again...

dirk-thomas:
Sorry, the latest change just added some "extra care" for Windows: https://github.com/ament/ament_tools/commit/ded89e7c1fa1dd0d24c26f72fa62ce966aa92ea2#diff-3e23bf3d285e5a32da12025ac88c215aR49

dirk-thomas:
> I don't see anything wrong with the changes, but I also don't clearly understand why d896162 fixes #109.

The refactoring commits to change anything. Only the third commit makes functional changes (introducing `get_setup_arguments_with_context`) which fixes the problem mentioned in the referenced ticket. 

> Is `get_setup_arguments` used anymore or is it meant to be completely replaced with `get_setup_arguments_with_context`?

`get_setup_arguments` is still being used. E.g. when a pure Python package is being handled. Since there are no dependencies declared the logic is being run with the current interpreter. The function is also being used from `get_setup_arguments_with_context`.

`get_setup_arguments_with_context` is being used when a package declared dependencies in its manifest. Then a custom environment including all build dependencies is generated and a separate Python interpreter is started to extract the information from the `setup()` call.

PR #150, opened 2017-06-20
pass exist_ok=True to makedirs calls

dirk-thomas:
Fix problems like this: http://ci.ros2.org/job/ci_windows/2917/console

[Build Status](http://ci.ros2.org/job/ci_windows/2918/)


dirk-thomas:
Since the CI build is passed the critical point I will go ahead and merge.

PR #156, opened 2017-08-07
remove arguments for specific script location

dirk-thomas:
Connect to ament/ament_python#3.

Required by ros2/demos#158 and ros2/examples#178.

### DIFF
--- a/ament_tools/build_types/ament_python.py
+++ b/ament_tools/build_types/ament_python.py
@@ -159,7 +159,6 @@ class AmentPythonBuildType(BuildType):
                 'build', '--build-base', os.path.join(
                     context.build_space, 'build'),
                 'install', '--prefix', context.install_space,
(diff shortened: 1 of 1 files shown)


dirk-thomas:
It seems like that on Windows `setuptools` uses a different default for `--install-scripts`: http://ci.ros2.org/job/ci_windows/3112

The options:
* Force every package to define `install-scripts` in the `setup.cfg` file with `$base/bin` (if not custom location is desired) to achieve cross platform compatibility.
* Modify `ament_tools` in the same way as it is planned for `bloom`: check for a `setup.cfg` file and its content and conditionally pass `--install-scripts` on Windows.

dirk-thomas:
As @wjwwood pointed out a third option is to add the default location (which is `<prefix>/Scripts`) to the `PATH` environment variable. I will try this first...

dirk-thomas:
See ament/ament_package#57.

PR #159, opened 2017-08-29
update style to satisfy new flake8 plugins

dirk-thomas:
Connect to ros2/ci#98.


ament_tools/build_type.py
```diff
@@ -53,7 +53,7 @@ class BuildAction:
     the optional ``env`` parameter.
     """
 
-    def __init__(self, cmd, type='command', title=None, dry_run_cmd=None,
+    def __init__(self, cmd, type='command', title=None, dry_run_cmd=None,  # noqa
```
mikaelarguedas: could you point out what the error we turn off here ? Would another wrapping strategy allow us to pass the test without ignoring the error?

ament_tools/build_type_discovery.py
```diff
@@ -42,10 +42,10 @@ def get_class_for_build_type(build_type):
     entry_points = list(yield_supported_build_types(build_type))
     if len(entry_points) > 1:
         # Shouldn't happen, defensive
-        raise RuntimeError("More than one build_type entry_point.")
+        raise RuntimeError('More than one build_type entry_point.')
     if len(entry_points) == 0:
         raise MissingPluginError(
             "No plugin to handle a package with build_type '{0}'"
-            .format(build_type)
+            .format(build_type),
```
wjwwood: Is this required? This doesn't make sense to me if `MissingPluginError` doesn't take any more arguments...

ament_tools/build_types/cmake.py
```diff
@@ -221,12 +221,12 @@ def _common_cmake_on_build(self, should_run_configure, context, prefix, extra_cm
                 cmd += msbuild_flags
                 # If there is a parallelism flag in msbuild_flags and it's not /m1,
                 # then turn on /MP for the compiler (intra-project parallelism)
-                if any([x.startswith('/m') for x in msbuild_flags]) and \
+                if any(x.startswith('/m') for x in msbuild_flags) and \
```
wjwwood: Is this really recommended? It seems to provide less information about what is happening and requires more expert knowledge to know that `any` is capable of this. -1 from me.

ament_tools/package_types/ament.py
```diff
@@ -19,10 +19,10 @@
...
-    parse_package=parse_package,
-    depends=[],
-)
+entry_point_data = {
+    'name': 'ament',
+    'description': "A package containing a '%s' manifest file." % PACKAGE_MANIFEST_FILENAME,
+    'package_exists_at': package_exists_at,
+    'parse_package': parse_package,
+    'depends': [],
+}
```
wjwwood: Was this part of the linter changes too? Is there a reason for this change? I actually think the previous syntax is more readable.

ament_tools/build_types/cmake.py
mikaelarguedas: Yeah I agree, this one is the only one that made me hesitate.
I think it's clearer that we have a list here if keeping the `[`

ament_tools/build_type.py
dirk-thomas: `type` is a built-in and shouldn't be used as a variable name.

ament_tools/build_type_discovery.py
dirk-thomas: We will not use the commas plugin: https://github.com/ros2/ci/pull/98#discussion_r135926777

PR #161, opened 2017-08-31
Fixed issue #160 by quoting executable's path

### LINKED ISSUE #160: Because of unquoted path ament fails on Windows
```
+++ Building 'ament_cppcheck'
'C:\Program' is not recognized as an internal or external command, operable program or batch file.
```
This happens because  ament_tools/ament_tools/setup_arguments.py contains the following command line construction code:

    # invoke get_setup_arguments() in a separate interpreter
    cmd = prefix + [sys.executable, '-c', '"%s"' % ';'.join(code_lines)]
    result = subprocess.run(
        ' '.join(cmd), stdout=subprocess.PIPE, shell=True, check=True)

If Python got installed into "Program Files" folder, the final  unquoted command line *cmd* can't be executed.

Proposed solution:
```
    sys_executable = sys.executable
    if not (sys_executable.startswith('"') and sys_executable.endswith('"')):
        sys_executable = '"' + sys_executable + '"'
    cmd = prefix + [sys_executable, '-c', '"%s"' % ';'.join(code_lines)]
```

serge-nikulin:
Fixes #160 
Connects to #160


mikaelarguedas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3125/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/482/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2491/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3180/)

dirk-thomas:
The triggered builds will only ensure that the use case without whitespaces in the path to the Python executable continues to work. Before merging this it should be ensured that the patch actually addresses the problem in that case.

mikaelarguedas:
Yes as discussed offline this set of jobs is only to ensure that it doesn't break the current state and I'll setup a windows machine with a python path with whitespaces and trigger another job to confirm this fixes the issue

dirk-thomas:
If additional directories (beside the Python executable path) contain spaces this patch isn't sufficient. I created #162 to quote all parts of the command correctly if necessary.

dirk-thomas:
Thank you for the patch. We merged #162 instead since it addresses the problem not only for the Python interpreter but for all pieces of the invoked command.

PR #162, opened 2017-08-31
shlex.quote shell command parts

dirk-thomas:
This patch makes sure that all parts of a shell command are quoted using `shlex.quote`. On Windows the function can't be used though since it quotes e.g. the executable with single quotes which doesn't work. Therefore on Windows a custom (very minimalistic) quoting is implemented.

It doesn't specifically target #160 but might fix it along the way. The quoting beyond #161 is still necessary in order to work for arguments which might also contain spaces.

Before (with whitespace in the various directories but not for the Python executable):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3127/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3178/)

After (with whitespace in the various directories but not for the Python executable):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3128/) (unrelated test failures)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3185/)

Just to make sure a set of "normal builds:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3129/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/484/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2493/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3186/)

### DIFF
--- a/ament_tools/helper.py
+++ b/ament_tools/helper.py
@@ -17,6 +17,7 @@ import filecmp
 from multiprocessing import cpu_count
 import os
 import re
+import shlex
 import shutil
 import stat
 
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
Another job with whitespaces in both the python executable path and the source/build/install paths
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3188/)

Looks like ci scripts need need a similar fix for this to work

dirk-thomas:
While the latest Windows job failed late in the process I think it still proves that this patch fixes the actual problem it is aiming to address.

mikaelarguedas:
Looks like the failure is due to rtiddsgen crashing sometimes when ran in parallel and is not related to this issue so I'd be in favor of merging this and close #160 and #161

PR #164, opened 2017-09-25
Start a bazel build tool.

clalancette:
This will allow us to use ament to build bazel projects.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>


ament_tools/build_types/bazel.py
```diff
@@ -0,0 +1,60 @@
...
+        return args, extras
+
+    def extend_context(self, options):
+        ce = ContextExtender()
+        ce.add('bazel_args', getattr(options, 'bazel_args', []))
+        return ce
+
+    def on_build(self, context):
+        yield BuildAction(['bazel', 'build'] + context.bazel_args + ['//...'],
+                          cwd=context.source_space)
```
dirk-thomas: Please don't align argument with the parenthesis.

Same below.

ament_tools/build_types/bazel.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def on_install(self, context):
+        # clalancette: Note that bazel has no real concept of an install
+        # target.  Thus, we define a de-facto one here which is a run
+        # command with a target of install.
```
dirk-thomas: Can you clarify what this target does?

ament_tools/build_types/bazel.py
```diff
@@ -0,0 +1,60 @@
...
+
+from ament_tools.build_type import BuildAction
+from ament_tools.build_type import BuildType
+
+from ament_tools.context import ContextExtender
+
+from ament_tools.helper import extract_argument_group
+
+
+class BazelBuildType(BuildType):
```
dirk-thomas: What about the not yet implemented method `on_uninstall` defined in the base class?

ament_tools/build_types/bazel.py
clalancette: One of the projects I want to use this with is drake.  in order to do an install with drake, you run:

```
bazel run //:install /path/to/install/target
```

The code in `on_install` is doing essentially the same thing; it calls the run command with a target of `//:install` .  Unfortunately, bazel has no concept of install, so there is no universal command or target that we can use.  We can either have a de-facto standard (i.e. we require all bazel projects that want to support ament have an install target to the run command), or we allow this to be configurable with more variables in the `<export/>` section (and allow the user to specify which command/target combination does an install).  Which of those to you prefer, or do you have another idea?

ament_tools/build_types/bazel.py
clalancette: I've changed this to be a separate variable for the cmd instead.

ament_tools/build_types/bazel.py
clalancette: I totally forgot, thanks.  I added this now.

clalancette:
> Can you please give a simple example which can be built locally using this.

Let me see if I can find a fairly simple bazel project for testing (all of the "real" projects I want to use this for have complicated setup/installation procedures).

ament_tools/build_types/bazel.py
dirk-thomas: Is there a way to check with bazel if the install target this code tries to call actually exists? If that is the case I think it would be fine (that is what the CMake build type does). If that is not possible it might be a good idea to split "plain bazel" support and "ament specific bazel" support (similar to CMake and ament cmake build types).

clalancette: It looks like there is a way to query bazel for whether a target exists, so I'll implement that concept.  Thanks.

clalancette:
> Let me see if I can find a fairly simple bazel project for testing (all of the "real" projects I want to use this for have complicated setup/installation procedures).

OK, I couldn't find a simple enough project, so I made one.  It's available here: https://github.com/clalancette/simple-bazel-install ; the README has instructions on how to build it.

ament_tools/build_types/bazel.py
```diff
@@ -0,0 +1,88 @@
...
+
+from ament_tools.context import ContextExtender
+
+from ament_tools.helper import extract_argument_group
+
+
+def _has_target(path, target):
+    # To query bazel for a list of all labels, we use:
+    # bazel query '//...' --output=label
+    output = subprocess.check_output(['bazel', 'query', '//...', '--output=label'], cwd=path)
```
dirk-thomas: I would suggest to define the `bazel` executable once as this package does for e.g. CMake and use the global variable everywhere: https://github.com/ament/ament_tools/blob/63ec41d7ff942295e52235bae07a91a13a439420/ament_tools/build_types/cmake_common.py#L21

ament_tools/build_types/bazel.py
```diff
@@ -0,0 +1,88 @@
...
+    def on_install(self, context):
+        # clalancette: Note that bazel has no real concept of an install
+        # target.  Thus, we define a de-facto one here which is a run
+        # command with a target of //:install.
+        if _has_target(context.source_space, '//:install'):
+            cmd = ['bazel', 'run'] + context.bazel_args + ['//:install', context.install_space]
+            yield BuildAction(cmd, cwd=context.source_space)
+        else:
+            self.warn("Could not install 'bazel' package because it has no "
+                      "'//:install' target")
```
dirk-thomas: Messages like this should include the actual package name to be more helpful for the user.

ament_tools/build_types/bazel.py
clalancette: Should be fixed in d4ca246

ament_tools/build_types/bazel.py
clalancette: Also fixed in d4ca246 .  I'll note that the cmake version of the messages does not do the correct thing here; I'll open another PR to address that.

ament_tools/build_types/bazel.py
```diff
@@ -0,0 +1,95 @@
...
+        cmd = [BAZEL_EXECUTABLE, 'build'] + context.bazel_args + ['//...']
+        yield BuildAction(cmd, cwd=context.source_space)
+
+    def on_install(self, context):
+        # clalancette: Note that bazel has no real concept of an install
+        # target.  Thus, we define a de-facto one here which is a run
+        # command with a target of //:install.
+        if _has_target(context.source_space, '//:install'):
+            cmd = [BAZEL_EXECUTABLE, 'run'] + context.bazel_args
+            cmd += ['//:install', context.install_space]
```
dirk-thomas: Rather than specifying the variable name again I would suggest:

```
cmd = [BAZEL_EXECUTABLE, 'run'] + context.bazel_args + \
    ['//:install', context.install_space]
```

Same below.

ament_tools/build_types/bazel.py
```diff
@@ -0,0 +1,95 @@
...
+            yield BuildAction(cmd, cwd=context.source_space)
+        else:
+            self.warn("Could not install '{0}' package because it has no "
```
dirk-thomas: I am not a native speaker but would "package '{0}'" sounds better?

clalancette: I think you are right that `package '{0}'` is slightly better, though both are perfectly understandable.  I'll reverse the order.

clalancette:
Thanks, squashing and merging.

PR #169, opened 2017-11-10
use pytest instead of nose

dirk-thomas:
Moving towards ros2/ros2#414.

Compare the "readability" of the test results to see the improvement (same test count but better grouping / hierarchy):

Builds before using `nose`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3494/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/726/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2829/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3544/)

Builds after using `pytest`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3495/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/727/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2825/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3548/)

Ready for review.


wjwwood:
Definitely easier to interpret the tests page on Jenkins.

mikaelarguedas:
looks much cleaner indeed.
Once we run the tests as pytest and not nosetests, do you think the grouping will also be better ? (instead of all being reported in `(root)`)

dirk-thomas:
> Once we run the tests as pytest and not nosetests, do you think the grouping will also be better ?

You mean switch from `ament_cmake_nose` to `ament_cmake_pytest`? Yes, I would expect that to "fix" more of the ugly grouping.

mikaelarguedas:
Yep that's what I meant. Cool :+1:

dhood:
Is anyone able to view the test results from pure python packages? I get a Proxy Error on e.g. http://ci.ros2.org/job/ci_linux/3495/testReport/demo_nodes_py.test/

dirk-thomas:
Changes in the wiki:
* [Developer-Guide](https://github.com/ros2/ros2/wiki/Developer-Guide/_compare/c9b96a39ca33cec8ff6753510a1e0f17e985586d...e879d471deb48ee8fdd7663c7b0e9dde5c9fb551)
* [Fedora-Development-Setup](https://github.com/ros2/ros2/wiki/Fedora-Development-Setup/_compare/f6d7fbd2b0dc0f8949e93fd0432ae06ea4c0b272...e879d471deb48ee8fdd7663c7b0e9dde5c9fb551)
* [Linux-Development-Setup](https://github.com/ros2/ros2/wiki/Linux-Development-Setup/_compare/8de1718c4ea01c0636f68d4b3dbfc1a24c69152e...e879d471deb48ee8fdd7663c7b0e9dde5c9fb551)
* [OSX-Development-Setup](https://github.com/ros2/ros2/wiki/OSX-Development-Setup/_compare/f2ebf9836b7f09c2e31fd04f625701334addafe4...e879d471deb48ee8fdd7663c7b0e9dde5c9fb551)
* [Windows-Development-Setup](https://github.com/ros2/ros2/wiki/Windows-Development-Setup/_compare/15a8a461a9953d8f87bdca39d7135d6decfc76ff...e879d471deb48ee8fdd7663c7b0e9dde5c9fb551)

mikaelarguedas:
> Is anyone able to view the test results from pure python packages?

hmm interesting I had no issue getting the `ament_flake8` test result. But when trying demo_nodes_py and a few other packages I get a proxy error as well

dirk-thomas:
> Is anyone able to view the test results from pure python packages?

> But when trying demo_nodes_py and a few other packages I get a proxy error as well

I just tried a minimal example and it seems I can open all the test results:

* Nose: http://ci.ros2.org/job/ci_linux/3505/
* Pytest: http://ci.ros2.org/job/ci_linux/3506/

mikaelarguedas:
> I just tried a minimal example and it seems I can open all the test results:

Me too :+1: but I still can see most python packages results from http://ci.ros2.org/job/ci_linux/3495.

Did you succeed to visualize the results of python packages from that job? 
http://ci.ros2.org/job/ci_linux/3495/testReport/ros2cli.test/
http://ci.ros2.org/job/ci_linux/3495/testReport/demo_nodes_py.test/
http://ci.ros2.org/job/ci_linux/3495/testReport/launch.test/

Maybe it comes from the current instability of ci.ros2.org but I'm afraid the result files are different / (bigger?) with pytest resulting in Jenkins having a hard time loading/displaying them. So we may want to confirm that before considering merging

dirk-thomas:
I noticed that the JUnit result files didn't contain any output. I reenabled capture in a3a958496c33354d5d2a8e35b1fd59a0dfe279f0. The following build shows the content (as it was before with `nose`): http://ci.ros2.org/job/ci_linux/3507/ / http://ci.ros2.org/job/ci_linux/3507/testReport/ament_package.test/test_copyright/test_copyright/

dirk-thomas:
> I'm afraid the result files are different / (bigger?) with pytest resulting in Jenkins having a hard time loading/displaying them

Can you reference two files which have that size difference? Then I can take a look at them.

mikaelarguedas:
Workspaces have been erased and I cant succeed to get jenkins to display the content so I have no idea if the diffeence is in the files themselves or in how jenkins display them.
I'll run a small build locally to see if anything is different in the result files.

dirk-thomas:
Locally the `launch/pytest.xunit.xml` file is 3.5k for me which seems pretty the size it was before.

mikaelarguedas:
Ok so we actually need pytest from pip3 for this to work out of the box. When using pytest-cov from apt I get the following error:
```
usage: setup.py [options] [file_or_dir] [file_or_dir] [...]
setup.py: error: argument --cov-report: invalid choice: 'xml:/home/mikael/work/ros2/test_sprint_ws/build_debug_isolated/launch/.coverage' (choose from 'term', 'term-missing', 'annotate', 'html', 'xml', '')
```

If I use all of them from pip I get files that are almost identical so I don't think the issue comes from the content of the files.

Not sure How the grouping work on jenkins so maybe the difference of behavior comes from the Jenkins plugin or because the master is pegget on CPU it cannot service requests for these folders...

dirk-thomas:
I can't reproduce the problem in other builds. I ran the `launch` tests since they are not opening in the old build:

* Nose: http://ci.ros2.org/job/ci_linux/3508/testReport/(root)/test_launch_with_coroutine/test_launch_with_coroutine/
* Pytest:  http://ci.ros2.org/job/ci_linux/3509/testReport/launch.test/test_launch_with_coroutine/test_launch_with_coroutine/

Both reporters use slightly different attribute in the xml but they both work fine except in the one build.

mikaelarguedas:
> Ok so we actually need pytest from pip3 for this to work out of the box. When using pytest-cov from apt I get the following error:

So I'm going to close https://github.com/ros2/ci/pull/110 can you update the linux and Fedora wiki pages accordingly?

mikaelarguedas:
> I can't reproduce the problem in other builds. I ran the launch tests since they are not opening in the old build:

Maybe that's because the master was in a weird state and that specific build ([3495](http://ci.ros2.org/job/ci_linux/3508)) was ran on the  `slave_on_master` node

dirk-thomas:
> So I'm going to close ros2/ci#110

Why not keep the `pytest` from Debian? Since you wanted it to be in-line with the doc.

> can you update the linux and Fedora wiki pages accordingly?

The version in Fedora should be fine. I updated the Ubuntu instructions: https://github.com/ros2/ros2/wiki/Linux-Development-Setup/_compare/3600486389fe0c5951fe0a25849f54a1807a2d9b...192005729672b3367479a03efde46dd008d35f41

dirk-thomas:
I triggered another full build: http://ci.ros2.org/job/ci_linux/3510/

mikaelarguedas:
> Why not keep the pytest from Debian? Since you wanted it to be in-line with the doc.

Because I have no way of testing it given that the 2 other packages installed via pip pull in the pip version of it.

dirk-thomas:
> Because I have no way of testing it given that the 2 other packages installed via pip pull in the pip version of it.

You could use `pip install --no-deps ...` :wink:

mikaelarguedas:
> You could use pip install --no-deps ... :wink:

Indeed I could do that but I'm not expecting our installation instructions do add `--no-deps` only for this package. That's why I'm suggesting to install them all via pip in both the instructions and ci until we can use the apt version for all of them (likely when we switch to bionic)

dirk-thomas:
For the latest full builds all the tests which couldn't be opened in the old build (while Jenkins was "troubled") are opening just fine:

* http://ci.ros2.org/job/ci_linux/3510/testReport/ros2cli.test/
* http://ci.ros2.org/job/ci_linux/3510/testReport/demo_nodes_py.test/
* http://ci.ros2.org/job/ci_linux/3510/testReport/launch.test/test_launch_with_coroutine/test_launch_with_coroutine/

Therefore I would consider this patch ready to be merged.

dirk-thomas:
I found another issue on Windows. The paths were not escaped so the JUnit files were not found by Jenkins. 19bcce757f162bbfe8633279e1de8cd89d006459 adds the necessary escaping which makes it work: http://ci.ros2.org/job/ci_windows/3564/testReport/launch.test/test_launch_with_coroutine/test_launch_with_coroutine/

PR #170, opened 2017-11-14
remove test_suite, add pytest as test_requires

mikaelarguedas:
connects to https://github.com/ros2/ros2/issues/414#issuecomment-344103560
Feel free to close if this was not changed on purpose


dirk-thomas:
Should the Travis file use `pytest` then instead of `nose`?

.travis.yml
```diff
@@ -28,6 +28,6 @@ install:
   - python setup.py install
   - cd ../..
 script:
-  - python setup.py nosetests -s
+  - python setup.py test
```
dirk-thomas: This doesn't work. See the latest Travis log which says "Ran 0 tests".

mikaelarguedas: This is working locally but not on travis. Will debug tomorrow and leave it in progress for now

mikaelarguedas:
Looks like pytest doesnt like test files with the same name in the same setup.py invocation.
I'm going to close this for now as I don't have time to look into it and that nose is fine with it

PR #171, opened 2017-11-15
generate PKG_CONFIG_PATH hook on demand

dirk-thomas:
As requested in ament/ament_cmake#113.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3542/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/749/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2885/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3623/)

### DIFF
--- a/ament_tools/build_types/cmake.py
+++ b/ament_tools/build_types/cmake.py
@@ -385,6 +385,12 @@ class CmakeBuildType(BuildType):
             environment_hooks_to_be_deployed.append(library_template_path)
             environment_hooks.append(os.path.join(environment_hooks_path, 'library_path.sh'))
 
+        # Prepare to deploy PKG_CONFIG_PATH environment hook
+        ext = '.sh' if not IS_WINDOWS else '.bat'
+        template_path = get_environment_hook_template_path('pkg_config_path' + ext)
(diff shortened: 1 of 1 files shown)


christianrauch:
Thanks for implementing this. It is adding the path to PKG_CONFIG_PATH.
However, it only does so after sourcing the workspace. The expanded PKG_CONFIG_PATH is not available during the initial build.

dirk-thomas:
> However, it only does so after sourcing the workspace. The expanded PKG_CONFIG_PATH is not available during the initial build.

@christianrauch Can you please elaborate the exact steps you are running, what you expect, and what the actual behavior is. Each package which installs `.pc` files should set or prepend to this environment variable when its setup file is being sources. This should happen whenever a package is being used as a build dependency by a downstream package.

dirk-thomas:
Based on the question in https://github.com/ament/ament_cmake/issues/113#issuecomment-345412372 and my [following answer](https://github.com/ament/ament_cmake/issues/113#issuecomment-348806071) I assume the new feature works well for your case and will merge this as-is.

PR #172, opened 2017-11-23
check CMAKE_COMMAND environment variable and look for cmake3 executab…

mallanmba:
Attempting to build ament on CentOS 7 with CMake 3.x fails because ament is hard coded to look for an executable named `cmake`. The RPM for CMake 3.x renames the executable `cmake3` to avoid conflict with CMake 2.x. This patch modifies ament_tools to check the CMAKE_COMMAND environment variable for user override and for a `cmake3` executable before looking for `cmake`

### DIFF
--- a/ament_tools/build_types/cmake.py
+++ b/ament_tools/build_types/cmake.py
@@ -24,8 +24,10 @@ from ament_tools.build_type import BuildAction
 from ament_tools.build_type import BuildType
 
 from ament_tools.build_types.cmake_common import CMAKE_EXECUTABLE
+from ament_tools.build_types.cmake_common import CMAKE_EXECUTABLE_ENV
 from ament_tools.build_types.cmake_common import cmakecache_exists_at
 from ament_tools.build_types.cmake_common import CTEST_EXECUTABLE
+from ament_tools.build_types.cmake_common import CTEST_EXECUTABLE_ENV
 from ament_tools.build_types.cmake_common import get_visual_studio_version
 from ament_tools.build_types.cmake_common import has_make_target
 from ament_tools.build_types.cmake_common import MAKE_EXECUTABLE
@@ -196,7 +198,8 @@ class CmakeBuildType(BuildType):
                 else:
                     cmake_args += ['-G', 'Unix Makefiles']
             if CMAKE_EXECUTABLE is None:
-                raise VerbExecutionError("Could not find 'cmake' executable")
+                raise VerbExecutionError("Could not find 'cmake' executable, try setting " +
+                                         CMAKE_EXECUTABLE_ENV)
             yield BuildAction(prefix + [CMAKE_EXECUTABLE] + cmake_args)
         elif IS_LINUX:  # Check for reconfigure if available.
             if MAKE_EXECUTABLE is None:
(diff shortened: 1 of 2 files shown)


gaoethan:
Does this means that we need to set CMAKE_COMMAND before building when there is version update of cmake ?  and the same goes for mac/windows?

mallanmba:
The CMAKE_COMMAND environment variable check is a way for a user to
explicitly tell ament what binary it should use. It is expected that this
would only be used by folks who are porting to a new platform or otherwise
doing something 'unusual'.
If CMAKE_COMMAND is unset (the common case), ament will check for an
executable named `cmake3` and if `cmake3` does not exist, it will try
`cmake`

mallanmba:
Is there a downside to checking for a `cmake3` binary, or is the request
for more verbose reporting of what ament is looking for?

`cmake3` is not an uncommon name for a CMake 3.x binary - the name was used
in Ubuntu Trusty, it's used in RedHat/CentOS, and it's quite likely to be
used elsewhere. Many of the 'enterprise' distributions are achingly slow to
update things and the cmake3 binary name is likely to stick around for
quite some time on platforms that still ship CMake 2.x by default. *If* a
cmake3 binary exists on a system, it'd be the most likely to satisfy the
CMake 3.5+ requirement of ament.

CMake find scripts typically handle these types naming or location
discrepancies when doing system introspection - it seems reasonable to me
that ament should have some flexibility/smarts when it's looking for the
right version of CMake. My first instinct was to check the version of
`cmake` binary and only check for `cmake3` if the version check failed, but
locating the binary and checking the version happen in different places in
ament.

Is there an objection to checking for `cmake3` or would more verbose error
checking be satisfactory?

tfoote:
As you've said you're making the assumption that the executable `cmake3` is preferred and more likely to be compliant than the `cmake` command just by it's being present.

If you want to make it more intellegent and check for `cmake3` if it determines that the `cmake` version is too low that could be reasonable. However the current logic would pick the `cmake3` version even if I have installed a custom version of cmake into my environment that is a high enough version.

And for example if you have a running system using the default `cmake` and then you install `cmake3` it will switch over, and you will have to set the `CMAKE_COMMAND` for it to return to the default command.

But really it comes down to being simple to understand, and to that end it seems that the behavior of using `cmake` unless `CMAKE_COMMAND` is set is the simplest and most expected behavior.

dirk-thomas:
Imo the approach in this PR makes sense. We also do exactly the same in other cases, e.g. finding [nose](https://github.com/ament/ament_cmake/blob/113915d0b8bd5e90e2649e9d39cf8cf5020a439f/ament_cmake_nose/ament_cmake_nose-extras.cmake#L25-L30).

Regaring the patch details I would suggest to replace the two new functions with a single one which takes the name of the environment variable and a list of executable names. That would avoid the code duplication. A minor point: please run the tests locally which should point out that some of the literal strings should use single quotes rather than double quotes.

mallanmba:
A single method to check for executable alternatives is used now. 

It still checks for `cmake3` before `cmake`, although I can see Tully's point that this wouldn't work for someone with a `cmake3` that is version 3.4 and a `cmake` in /usr/local that is version 3.6. I'm happy to change the search order to put cmake first if that's strongly desired. 

If the cmake version check fails, is there a way to catch that and print a help message? It wasn't immediately obvious to me how to do that.

dirk-thomas:
Thank you for contributing this improvement.

PR #176, opened 2018-02-13
Pandering to finicky CMake Initial Cache ordering

stonier:
Problem: https://github.com/ament/ament_cmake/issues/122
Fix: makes sure the CMake args get added after other args and puts the source path at the end

Not 100% sure if this is bulletproof - could use an experienced set of eyes on this.


dirk-thomas:
Thank you for the patch. Adding the source space at the very end seems like the right approach.

But I don't think the other parts (extra CMake args, install prefix, generator) should / need to be changed in order. The rational for the order is that the passed extra CMake args can't pass an install prefix or generator. Therefore those are always set after the user provided extra CMake args.

Can you please update the patch to only move the source space to the end and confirm that this also fixes your problem.

stonier:
Aye, the source path is the only requirement. Works without having to shift the cmake args.

ament_tools/build_types/cmake.py
```diff
@@ -175,8 +175,7 @@ def _common_cmake_on_build(self, should_run_configure, context, prefix, extra_cm
         # Execute the configure step
         # (either cmake or the cmake_check_build_system make target)
         if should_run_configure:
-            cmake_args = [context.source_space]
-            cmake_args.extend(extra_cmake_args)
+            cmake_args = extra_cmake_args
```
dirk-thomas: This will mutate the passed in parameter `extra_cmake_args` which is undesired. Please ensure to duplicate the list, e.g. using `cmake_args = list(extra_cmake_args)`.

stonier: Oh, noob. Showing signs of pythonic rust. Fixed and tested.

dirk-thomas:
Thank you for the patch.

PR #178, opened 2018-02-17
separate loops for evaluating conditions and extracting group members

dirk-thomas:
This patch was missing from ament/ament_package#69 and why it worked locally but not on the farm.


mikaelarguedas:
topological order now passes locally with these 2 PRs

PR #184, opened 2018-03-27
turn on BUILD_TESTING for cmake packages too, if --build-tests is passed

wjwwood:
I noticed, while working on a plain cmake package for our testing tools, that `BUILD_TESTING` is not turned on for plain cmake package, only for `ament_cmake` packages.

I'm not sure if that was intentional, but it was turned on for `ament_cmake` packages in this pr:

https://github.com/ament/ament_tools/pull/92

Which was part of the large shift to that from our own testing variable:

https://github.com/ament/ament_cmake/pull/68

Anyways, it's annoying if I want to test plain cmake packages to have to pass them arguments (even if I were able to narrow it down). This seems like reasonable behavior to me.

It might cause some new tests to appear on our CI, however, but we'll see:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4148/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1225/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3429/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4252/) (false positive cmake warning due to interleaved output)


ament_tools/build_types/cmake.py
```diff
@@ -143,6 +143,7 @@ def on_build(self, context):
         # Calculate any extra cmake args which are not common between cmake build types
         extra_cmake_args = []
         if should_run_configure:
+            extra_cmake_args += ['-DBUILD_TESTING=%d' % int(context.build_tests)]
```
dirk-thomas: Shouldn't this be outside the `should_run_configure` condition? Otherwise it wouldn't change when the `--build-tests` is toggled.

wjwwood: Not sure, I was copying `ament_cmake`:

https://github.com/ament/ament_tools/blob/6294c63fdd2a55bc4a48ffcf937cb0b6a0fb19d9/ament_tools/build_types/ament_cmake.py#L98-L99

As I read it, the content of the `extra_cmake_args` are only important if `should_run_configure` is already true. Moving this out of that if conditional shouldn't change the behavior I think. As to why it should not be added to `extra_cmake_args` if configure won't be done, I don't know.

wjwwood: The triggering of cmake is already based on `--build-tests` here:

https://github.com/ament/ament_tools/blob/6294c63fdd2a55bc4a48ffcf937cb0b6a0fb19d9/ament_tools/build_types/ament_cmake.py#L84

Maybe I don't understand what you're asking though.

dirk-thomas: Ok, I see that the same check is already in place for plain CMake packages: https://github.com/ament/ament_tools/pull/184/files#diff-9e94d1bd2a8b57f7c156534b14d8ad3fR131

Sounds good to me than as-is.

Issue #185, opened 2018-04-18
python3-pytest doesn't support option "-o" and it results in error while running python unittest of a ros2 component


gaoethan:
What `python3-pytest` I installed is `2.8.7-4` and there is no support to option `-o` now from the checking of output via:
`python3 -u -m pytest --help`

After enabling python unittest of a ROS2 components, for example:
```
find_package(ament_cmake_pytest REQUIRED)
ament_add_pytest_test(enumerants.py "enumerants.py")
```
then build and run the tests accordingly
`ament test --skip-build --skip-install --only-packages <your-component>`

The following error will occurs:
```
2: Test timeout computed to be: 60
2: -- run_test.py: invoking following command in '/home/ethan/Myspace/rosdev/2upstream/vision_opencv/cv_bridge':
2:  - /usr/bin/python3 -u -m pytest /home/ethan/Myspace/rosdev/2upstream/ci/src/cv_bridge/test/enumerants.py -o cache_dir=/home/ethan/Myspace/rosdev/2upstream/ci/build/cv_bridge/test/ament_cmake_pytest/enumerants.py/.cache --junit-xml=/home/ethan/Myspace/rosdev/2upstream/ci/build/cv_bridge/test_results/cv_bridge/enumerants.py.xunit.xml --junit-prefix=cv_bridge
2: usage: pytest.py [options] [file_or_dir] [file_or_dir] [...]
2: pytest.py: error: unrecognized arguments: -o cache_dir=/home/ethan/Myspace/rosdev/2upstream/ci/build/cv_bridge/test/ament_cmake_pytest/enumerants.py/.cache
2:   inifile: None
2:   rootdir: /home/ethan/Myspace/rosdev/2upstream
2: -- run_test.py: return code 2
2: -- run_test.py: generate result file '/home/ethan/Myspace/rosdev/2upstream/ci/build/cv_bridge/test_results/cv_bridge/enumerants.py.xunit.xml' with failed test
2: -- run_test.py: verify result file '/home/ethan/Myspace/rosdev/2upstream/ci/build/cv_bridge/test_results/cv_bridge/enumerants.py.xunit.xml'
2/4 Test #2: enumerants.py ....................***Failed    0.48 sec
```

mikaelarguedas:
Did you try installing pytest from pip as specified in the [installation instructions](https://github.com/ros2/ros2/wiki/Linux-Development-Setup) ?

gaoethan:
yes. I've already installed as that procedure.
[...]

PR #186, opened 2018-04-18
Fix python unittest runtime error in ament_tools

gaoethan:
Fix python unittest runtime error due to no support to option `-o` in `python3-pytest`

Signed-off-by: Ethan Gao <ethan.gao@linux.intel.com>


gaoethan:
This is the part of the hot patch to issue #185

gaoethan:
No need to fix with this PR after upgrading `pytest` to latest e.g **3.5.0**, abandon it !

PR #187, opened 2018-05-23
Allow RelWithDebInfo and MinSizeRel configurations on cmake

MiguelCompany:
I had the need to build ROS2 on MSVS using RelWithDebInfo configuration, and discovered that ament_tools changed the value passed with --cmake-args -DCMAKE_BUILD_TYPE=RelWithDebInfo to 'Release'

This PR addresses this case


dirk-thomas:
Thank you for the patch.

I fixed the style to pass the linters / tests.

PR #189, opened 2019-11-01
Install package.xml

galou:
This avoids a warning about future non-supported feature with colcon.


dirk-thomas:
This package is EOLed and isn't used anymore since `colcon` fulfills it's role. Therefore I am not sure why it should be updated.

We should probably consider archiving this repo...

galou:
OK, thanks for the info.

